PR #52, opened 2026-06-25
fix(minimal-template): remove phantom `memory search` / `route` CLI commands from docs (#48)

### LINKED ISSUE #48: bug: minimal template CLAUDE.md documents `memory search` / `route` commands that bin/cli.js never implements
## Summary

A harness generated from the **`minimal`** template ships a `CLAUDE.md` that documents two CLI commands the generated `bin/cli.js` never implements:

| Documented in `CLAUDE.md` | Reality |
|---|---|
| `<name> memory search <query>` — "Semantic search across stored patterns" | Not implemented — falls through to the `default:` case |
| `<name> route <task>` — "Get the routing tier recommendation" | Not implemented — falls through to the `default:` case |

Running either prints `Unknown command: memory` / `Unknown command: route` and exits `2`. The behavioral-rules section reinforces the false promise ("Memory and routing are handled by the kernel"), so an agent reading `CLAUDE.md` will confidently invoke commands that always error.

## Steps to reproduce

1. `npx create-agent-harness my-harness` (select the **minimal** template)
2. `cd my-harness && npm install`
3. `node ./bin/cli.js memory search foo`

Observed:

```
Unknown command: memory. Try `my-harness --help`.
```

(exit code `2`). Same for `route`.

## Root cause

`templates/minimal/CLAUDE.md.tmpl` advertises the commands:

```md
| `{{name}} memory search <query>` | Semantic search across stored patterns |
| `{{name}} route <task>`          | Get the routing tier recommendation |
```

…but `templates/minimal/bin/cli.js.tmpl`'s dispatch switch only handles `init`, `doctor`, `--version`/`-v`, `--help`/`-h`. There is no `case 'memory'` / `case 'route'`, so both hit:

```js
default:
  console.error(`Unknown command: ${cmd}. Try \`${HARNESS_NAME} --help\`.`);
  return 2;
```

## Scope

- The phantom command claims live **only** in `templates/minimal/CLAUDE.md.tmpl` (verified — no other template's docs make this claim).
- **No** template implements a `memory` or `route` dispatch case (verified — `grep "case '(memory|route)'"` across `templates/` returns nothing).

So this is isolated to the `minimal` template's docs ↔ code fidelity.

## Note on whether the capability exists

For context (not prescribing a fix) — the published `@metaharness/kernel` surface doesn't make these turnkey:

- **memory**: `@metaharness/kernel/memory` exports `rankWithDecay(hits, config)`, `distillTrigger()`, `linearDecay()`. These *re-rank hits already retrieved*; the kernel source notes the actual vector store is AgentDB-backed. There's no string-query → embed → search → results function shipped, so `memory search <query>` is more than a thin wrapper.
- **route**: the kernel README lists a `routing` subsystem, but the package `exports` map only publishes `.`, `./memory`, `./memory-rvf`, `./dispatch` — no routing entry point. So `route` can't currently be implemented from the package as published.

## Possible solutions (maintainer's call)

1. **Trim the docs to match the code** — remove the two rows (and soften the "Memory and routing are handled by the kernel" line) from `templates/minimal/CLAUDE.md.tmpl`. Lowest-risk; makes the scaffold honest immediately.
2. **Implement the commands** — wire `case 'memory'` (AgentDB retrieval + `rankWithDecay`) and, once the kernel exposes routing, `case 'route'`, in `bin/cli.js.tmpl`. Higher effort; `route` is blocked on the kernel not exporting a routing entry point.
3. **Stub with a clear "not yet implemented" message** — make the commands exist but exit non-zero with a pointer to the roadmap, so docs and CLI agree without over-promising capability.

Happy to send a PR for whichever direction you prefer.

## Environment

- Generated harness from the `minimal` template
- Node.js >= 20
- Reproducible cross-platform (docs/dispatch mismatch is OS-independent)

---

Related: #44 (same class of scaffold-fidelity defect in the generated output).

ruvnet:
Fixes #48 (thanks @Tovli for the precise report).

The `minimal` template's `CLAUDE.md.tmpl` documented two CLI commands — `memory search <query>` and `route <task>` — that `bin/cli.js.tmpl` never implements; both fall through to the `default:` `Unknown command` branch (exit 2). An agent reading the generated CLAUDE.md would confidently invoke commands that always error.

As the issue notes, the published `@metaharness/kernel` surface doesn't make these turnkey (memory exposes `rankWithDecay`/`distillTrigger` re-rankers, not a string→embed→search; routing is a library subsystem). So the honest fix is **docs↔code fidelity**, not faking the commands:

- Removed the two phantom command rows; the table now lists exactly what the CLI implements (`init`, `doctor`, `--version`, `--help`) — matching cli.js's own `--help` text.
- Reworded the behavioral-rules line so it no longer implies memory/routing are turnkey — it now says the kernel ships them as **library primitives**, call them from your own code.

Scope: one file, `templates/minimal/CLAUDE.md.tmpl`. No code/behavior change. Verified (per the issue) that no template implements a `memory`/`route` dispatch and only the minimal docs made the claim.

### DIFF
--- a/packages/create-agent-harness/templates/minimal/CLAUDE.md.tmpl
+++ b/packages/create-agent-harness/templates/minimal/CLAUDE.md.tmpl
@@ -5,18 +5,19 @@
 ## Behavioral rules
 
 - Use the harness's MCP tools (`mcp__{{name}}__*`) for orchestration
-- Memory and routing are handled by the kernel — you don't need to learn them
+- The kernel ships memory-ranking and routing **primitives** as a library (`@metaharness/kernel`); the minimal template does not expose them as CLI commands — call them from your own code if needed
 - Defer destructive operations to the user
 
 ## Commands
 
-After `{{name}} init`, the following are available:
+The minimal template's CLI (`bin/cli.js`) implements exactly these:
 
 | Command | What it does |
 |---|---|
-| `{{name}} doctor` | Health check the install |
-| `{{name}} memory search <query>` | Semantic search across stored patterns |
-| `{{name}} route <task>` | Get the routing tier recommendation |
+| `{{name}} init` | Boot the kernel + host adapter (default command) |
+| `{{name}} doctor` | Health-check the install end-to-end |
+| `{{name}} --version` | Print the kernel version |
+| `{{name}} --help` | Show usage |
 
 ## Architecture
 


ruvnet:
Superseded by #161 — the underlying issue (#48) was already fixed on `main` a different way while this PR sat stale (conflicting, and CI failing on an unrelated pre-existing `scripts/gcp-cluster.mjs` lint issue that's since been fixed on `main`). #161 folds in the one still-missing piece (documenting `--version`) on top of current main.

PR #53, opened 2026-06-25
feat: add Requesty as an OpenAI-compatible mutator

Thibaultjaigu:
Adds `RequestyMutator`, an optional LLM-backed `CodeGenerator` for Darwin Mode, mirroring the existing `OpenRouterMutator` 1:1. Requesty is an OpenAI-compatible gateway.

## What
- New file `packages/darwin-mode/src/requesty-mutator.ts` — a verbatim mirror of `openrouter-mutator.ts`, changing only:
  - class `OpenRouterMutator` → `RequestyMutator`, `OpenRouterMutatorOptions` → `RequestyMutatorOptions`
  - key read from `REQUESTY_API_KEY` (env, with a `/tmp/.rqkey` dev fallback, same pattern as the sibling)
  - endpoint `https://router.requesty.ai/v1/chat/completions`
  - default model `openai/gpt-4o-mini` (provider/model naming matches OpenRouter)
  - summary string prefix `requesty:${model} ...`
  - imports `MutatorTelemetry` from `openrouter-mutator.js` rather than re-declaring it, so `export *` from both files does not collide
- `packages/darwin-mode/src/index.ts` — `export * from './requesty-mutator.js';` beside the OpenRouter export, plus a mirrored doc-comment line

It slots in behind the SAME `validateGeneratedCode` safety gate as the other generators — pure refactor, no new capabilities.

## Tested
- `npx tsc --noEmit` (the package `lint` script) passes clean.
- Live request to `https://router.requesty.ai/v1/chat/completions` with `model: openai/gpt-4o-mini`, `max_tokens: 32` → HTTP 200 with a real completion.

I work at Requesty. This mirrors the existing OpenRouter provider as closely as possible. Happy to adjust wording/placement or close it if it's not a fit.

### DIFF
--- a/packages/darwin-mode/src/index.ts
+++ b/packages/darwin-mode/src/index.ts
@@ -18,6 +18,8 @@
 //   generator      — RepoProfile → baseline variant
 //   mutator        — bounded, validated child mutations + the CodeGenerator hook
 //   openrouter-mutator — optional LLM-backed CodeGenerator (same safety gate)
+//   requesty-mutator   — optional LLM-backed CodeGenerator (Requesty is an
+//                        OpenAI-compatible gateway; same safety gate)
 //   sandbox        — gate-first, shell-free, env-scrubbed task runner
 //   scorer         — the frozen weighted scorer + strict promotion gate (ADR-072)
 //   archive        — the population tree + archive-wide selection (ADR-073)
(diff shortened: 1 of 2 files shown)


ruvnet (approved):
Thank you, @Thibaultjaigu! 🙏 This is a clean, faithful mirror of the OpenRouter mutator — and it even tidies a cross-platform rough edge (`os.tmpdir()` instead of the sibling's hardcoded `/tmp`). The type-only `MutatorTelemetry` import to avoid the `export *` collision was exactly the right call, and the network-failure→safe-no-op keeps it behind the same `validateGeneratedCode` safety gate with no new capabilities. tsc clean + live-tested. Appreciate the transparency about working at Requesty, and the care to match the existing pattern precisely. Merging — welcome aboard! 🚀

PR #102, opened 2026-07-06
feat(flywheel): per-generation checkpoint hook — crash-recoverable long runs (0.1.4)

ruvnet:
## What
Adds an optional `onGeneration` checkpoint hook to `runFlywheelGenerations` (@metaharness/flywheel 0.1.3 → 0.1.4): after each generation it emits a **complete, independently replay-verifiable** `ReplayBundle` for the run so far (`GenerationCheckpoint { generation, generationsRun, partialBundle, spent }`). `d1s4-live-run.mjs` persists it to `proof-bundle-swebench.partial.json` each generation.

## Why
The D1 positive-path (agentic) run spans **hours**; the engine only wrote the bundle at the very end, so a crash/OOM/rate-limit lost everything (the prior single-shot run crashed at ~45min with nothing saved). Now a crash keeps the completed generations' evidence — directly de-risks the flagship D1 run.

## Safety
- **Observation-only**: the hook never influences promotion; a **throwing hook is swallowed** and can't abort a valid run (tested).
- `meetsPromotionRule` **untouched**; the final bundle + each checkpoint are assembled by ONE shared `buildBundle` closure so they can't diverge.

## Tests
flywheel **21/21** (+2: each per-gen checkpoint independently replay-verifies; a throwing hook can't kill the run). The existing acceptance test confirms the `buildBundle` DRY refactor is **behavior-identical**. tsc clean. ADR-236 updated.


aegntic:
this is exactly what was missing

PR #151, opened 2026-07-28
fix(proxy): bump META_PROXY_VERSION → 0.7.0 + port pin-drift watcher (#149)

### LINKED ISSUE #149: META_PROXY_VERSION pinned at 0.4.1 — three meta-proxy releases behind, and the pin-drift watcher guards the wrong repo
**Summary.** `META_PROXY_VERSION` in `packages/create-agent-harness/src/meta-proxy.ts` is `0.4.1`, but the public install mirror `cognitum-one/meta-proxy-dist` has published `v0.5.0`, `v0.6.0` and `v0.7.0`. Because `metaharness proxy install` downloads the **pinned** version, every user running `npx metaharness proxy install --yes` today gets a proxy three releases old and is told it succeeded:

```
Installed verified Meta-Proxy v0.4.1 at ~/.metaharness/meta-proxy/bin/meta-proxy.
```

**Why now.** `v0.4.1` was published 2026-07-17 and the pin was correct at publish time. Three meta-proxy releases then shipped on 2026-07-27 and nothing in this repo watches for that. Nobody gets the newer proxy until someone remembers to hand-edit this constant, and the install output gives no hint the pin is stale.

A pin-drift watcher for this exact failure mode already exists — but it was built in **cognitum-one/metaharness**, guarding `PINNED_PROXY_VERSION` in that repo's `packages/cli/src/proxy-install.ts`. That is not the package users install. `npx metaharness` resolves to the npm package published from **this** repo, whose pin has no watcher at all. The guardrail is pointed at the wrong constant.

**Done looks like.** `npx metaharness proxy install --yes` installs the latest signed meta-proxy release, and CI notices within a day when a new release lands and this pin doesn't follow.

**Work items.**
1. Confirm `PROXY_SIGNING_PUBKEY_PEM` still matches meta-proxy's `signing-key.pub.pem` (re-pin in the same change if the key rotated), then bump `META_PROXY_VERSION` to `0.7.0`.
2. Port the pin-drift check to this repo — a scheduled job diffing `META_PROXY_VERSION` against `meta-proxy-dist`'s latest release tag, opening/closing a tracking issue. Reference implementation: `cognitum-one/metaharness/.github/workflows/proxy-pin-drift.yml`.
3. Publish the bumped CLI; verify one real `proxy install` on macOS + Linux.

---

### Evidence

Pin, read from the published tarball (`npm pack metaharness@0.4.1`), so this is what users actually run — not just what's on `main`:

```
package/dist/meta-proxy.js:15:export const META_PROXY_VERSION = '0.4.1';
package/dist/meta-proxy.js:16:export const META_PROXY_RELEASE_BASE = 'https://github.com/cognitum-one/meta-proxy-dist/releases/download';
```

`cognitum-one/meta-proxy-dist` releases:

| tag | published |
|---|---|
| v0.7.0 | 2026-07-27 (latest) |
| v0.6.0 | 2026-07-27 |
| v0.5.0 | 2026-07-27 |
| v0.4.1 | 2026-07-17 ← pinned |

`main` here still reads `0.4.1` (`packages/create-agent-harness/src/meta-proxy.ts:29`), so this is live drift, not an unpublished fix.

### Two-repo split worth naming

The `metaharness` name spans two repos and the guardrails landed asymmetrically:

| | `ruvnet/metaharness` (this repo) | `cognitum-one/metaharness` |
|---|---|---|
| ships as | `npx metaharness` on npm | private/internal CLI |
| pin constant | `META_PROXY_VERSION` = `0.4.1` | `PINNED_PROXY_VERSION` = `0.1.0` |
| drift watcher | **none** | `proxy-pin-drift.yml` (tracking issue open) |

Both pins are stale; only one is watched, and it's the one users don't install. Work item 2 closes that gap. A single shared watcher config across the two repos would be better than two copies, if there's appetite for it.

### Note: bumping this pin does not fix the other reported failure

The same session also hit this on `proxy login`:

```
meta-proxy: /lib/aarch64-linux-gnu/libc.so.6: version `GLIBC_2.39' not found (required by .../meta-proxy)
```

That is a separate, independent defect in the meta-proxy release pipeline — filed as cognitum-one/meta-proxy#65. Verified: the `v0.7.0` artifact carries the identical `GLIBC_2.39` floor as `v0.4.1`, so bumping this pin will move users from a broken-on-old-glibc `0.4.1` to a broken-on-old-glibc `0.7.0`. Sequence the bump after #65 lands, or the pin bump will read as a fix that didn't fix anything.

ruvnet:
Closes #149.

`metaharness proxy install` downloads the **pinned** meta-proxy version. The pin sat at `0.4.1` (2026-07-17) while `cognitum-one/meta-proxy-dist` shipped v0.5.0/0.6.0/0.7.0 on 2026-07-27 — every `npx metaharness proxy install --yes` got a proxy three releases old and was told it succeeded. The existing watcher guarded the wrong repo's constant.

## Changes
- **Bump** `META_PROXY_VERSION` `0.4.1` → `0.7.0` (`packages/create-agent-harness/src/meta-proxy.ts`).
- **Port the watcher**: `.github/workflows/proxy-pin-drift.yml` — scheduled diff of `META_PROXY_VERSION` vs `meta-proxy-dist`'s latest release tag, opening/updating a tracking issue on drift, pointed at **this** repo's constant (the one `npx metaharness` actually installs).

## ⚠ Merge checklist (I could not verify these — signing material is not mine to read)
- [ ] Confirm `META_PROXY_SIGNING_PUBKEY_PEM` still matches meta-proxy's `signing-key.pub.pem` for **v0.7.0**. If the key rotated, re-pin the PEM in this same change — otherwise install-time signature verification breaks.
- [ ] Confirm `v0.7.0` is still the latest signed `meta-proxy-dist` release at merge time.
- [ ] Verify one real `metaharness proxy install --yes` on macOS + Linux.

🤖 Generated with [RuFlo](https://github.com/ruvnet/ruflo)

### DIFF
--- a/.github/workflows/proxy-pin-drift.yml
+++ b/.github/workflows/proxy-pin-drift.yml
@@ -1,0 +1,50 @@
+# proxy-pin-drift — watch META_PROXY_VERSION against meta-proxy-dist releases.
+#
+# The pin lives in THIS repo (the package users install via `npx metaharness`);
+# the pre-existing watcher was built in cognitum-one/metaharness guarding a
+# different constant, so drift here went unnoticed (#149). This ports the check
+# to the repo that actually publishes the pin.
+name: proxy-pin-drift
+
+on:
+  schedule:
+    - cron: '23 5 * * 1'   # Mondays 05:23 UTC (off-peak minute)
+  workflow_dispatch:
+
+permissions:
+  contents: read
+  issues: write
+
+jobs:
+  drift:
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - uses: actions/checkout@v4
+      - name: Compare META_PROXY_VERSION to meta-proxy-dist latest release
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          PINNED=$(grep -oP "META_PROXY_VERSION = '\K[^']+" packages/create-agent-harness/src/meta-proxy.ts)
+          LATEST=$(gh release list --repo cognitum-one/meta-proxy-dist --limit 1 --json tagName --jq '.[0].tagName' | sed 's/^v//')
+          echo "pinned=$PINNED  latest=$LATEST"
+          if [ -z "$LATEST" ]; then echo "could not read meta-proxy-dist latest release — skipping (not a drift)"; exit 0; fi
+          if [ "$PINNED" = "$LATEST" ]; then echo "✓ META_PROXY_VERSION is current."; exit 0; fi
+          TITLE="meta-proxy pin drift — META_PROXY_VERSION ($PINNED) is behind meta-proxy-dist v$LATEST"
+          BODY_FILE=/tmp/body.md
+          {
+            echo "\`META_PROXY_VERSION\` is pinned at \`$PINNED\` but \`cognitum-one/meta-proxy-dist\`"
+            echo "latest release is \`v$LATEST\`. \`metaharness proxy install\` downloads the pinned"
+            echo "version, so users are getting a stale proxy and being told it succeeded."
+            echo ""
+            echo "**To resolve:** confirm \`META_PROXY_SIGNING_PUBKEY_PEM\` still matches meta-proxy's"
+            echo "\`signing-key.pub.pem\` for \`v$LATEST\` (re-pin it in the same change if the key"
+            echo "rotated), then bump \`META_PROXY_VERSION\` in"
+            echo "\`packages/create-agent-harness/src/meta-proxy.ts\` and publish the CLI."
+          } > "$BODY_FILE"
+          EXISTING=$(gh issue list --repo "$GITHUB_REPOSITORY" --search "$TITLE in:title" --state open --json number --jq '.[0].number' 2>/dev/null || true)
+          if [ -n "$EXISTING" ]; then
+            gh issue comment "$EXISTING" --repo "$GITHUB_REPOSITORY" --body-file "$BODY_FILE"
+          else
+            gh issue create --repo "$GITHUB_REPOSITORY" --title "$TITLE" --label bug --body-file "$BODY_FILE"
+          fi
--- a/packages/create-agent-harness/src/meta-proxy.ts
+++ b/packages/create-agent-harness/src/meta-proxy.ts
@@ -26,7 +26,7 @@ import {
 import { homedir } from 'node:os';
 import { basename, join } from 'node:path';
 
-export const META_PROXY_VERSION = '0.7.0';
+export const META_PROXY_VERSION = '0.7.1';
 export const META_PROXY_RELEASE_BASE = 'https://github.com/cognitum-one/meta-proxy-dist/releases/download';
 
 /** The release signing key, pinned in the client rather than fetched from GitHub. */
--- a/packages/create-agent-harness/src/meta-proxy.ts
+++ b/packages/create-agent-harness/src/meta-proxy.ts
@@ -26,7 +26,7 @@ import {
 import { homedir } from 'node:os';
 import { basename, join } from 'node:path';
 
-export const META_PROXY_VERSION = '0.4.1';
+export const META_PROXY_VERSION = '0.7.0';
 export const META_PROXY_RELEASE_BASE = 'https://github.com/cognitum-one/meta-proxy-dist/releases/download';
 
 /** The release signing key, pinned in the client rather than fetched from GitHub. */


proffesor-for-testing:
Reviewed. **This does solve #149** — both work items are covered (bump + watcher, ported to the repo whose pin `npx metaharness` actually installs). Two things before merging, one of them blocking.

## ✅ Checklist items I could verify for you

**Signing key — no rotation needed, and I proved it rather than string-matching.** `META_PROXY_SIGNING_PUBKEY_PEM` on this branch is byte-identical to meta-proxy's `signing-key.pub.pem`, and I ran the real `v0.7.0` release material through it:

```
$ node scripts/verify-release-signature.mjs SHA256SUMS SHA256SUMS.sig signing-key.pub.pem
release_signature=valid public_key_sha256=a0fec4760b81b2819318c17707b02846b88025043a9bccd16dfcdbe37012ddbd
```

**`v0.7.0` is currently `isLatest` in `meta-proxy-dist`, with all five platform assets + `SHA256SUMS` + `SHA256SUMS.sig` present.** So the pin is installable as written.

Also checked: `dist/` is gitignored and built by `tsc` at publish, so the `src/` bump does reach the published package — no stale-`dist` trap. The `bug` label exists here, so `gh issue create --label bug` won't 422. `grep -oP` matches correctly on GNU grep (tested on `ubuntu:24.04` against this branch's file → `0.7.0`).

## 🔴 Blocking: `0.7.0` is the wrong pin target as of today

`v0.7.0` carries a **`GLIBC_2.39` floor** and cannot start on Debian 12 (the default devcontainer/devpod base, glibc 2.36), Ubuntu 22.04, or RHEL 9 / Amazon Linux 2023 — it dies in the dynamic loader before `main`, so every subcommand fails identically. That is cognitum-one/meta-proxy#65, and it is what the user in #149 actually hit; the stale pin and the loader error were two independent defects.

I verified `v0.7.0` fails *identically* to `v0.4.1` here, so merging as-is moves users from a broken `0.4.1` to a broken `0.7.0` — the stale-pin half of #149 is fixed, but the install stays unusable on the most common environment.

The fix merged in cognitum-one/meta-proxy#66 and ships as **`v0.7.1`** (cognitum-one/meta-proxy#67, green, mirroring to `meta-proxy-dist` is automated in that same PR). **Suggest retargeting this pin to `0.7.1`** once it lands. Happy to push that one-line change to this branch if you'd like.

## 🟡 Non-blocking: the watcher will accumulate duplicate issues

The dedup search is fine syntactically — I confirmed `--search "<title> in:title"` matches an existing issue even with the em-dash (it found #149). The problem is that the title **embeds both version numbers**:

```
meta-proxy pin drift — META_PROXY_VERSION ($PINNED) is behind meta-proxy-dist v$LATEST
```

So the moment either side moves, the search looks for a title that doesn't exist yet, misses the still-open issue, and files a second one. Concretely, this is about to happen: it opens `…(0.4.1) is behind …v0.7.0`, we publish `0.7.1`, and the next run opens `…(0.4.1) is behind …v0.7.1` while the first stays open forever. There's also no auto-close when the pin catches up.

`cognitum-one/metaharness/.github/workflows/proxy-pin-drift.yml` matches on a stable title **prefix** and `gh issue close`s when `PINNED == LATEST` for exactly this reason — worth lifting those two behaviours.

Two smaller ones:
- **Cadence.** Weekly (Mondays 05:23) means up to a 7-day window where the pin is stale and nothing says so. The cognitum one runs daily. Given `0.7.1` lands this week, weekly means the drift notice could sit until next Monday.
- **`gh release list --limit 1`** returns the newest release *by date*, which is not necessarily the latest stable — a draft or prerelease would win. It returns `v0.7.0` correctly today, but `gh api repos/cognitum-one/meta-proxy-dist/releases/latest` (what the cognitum workflow uses) excludes drafts and prereleases by definition.

Nothing here blocks the watcher being a clear improvement over having none — these are follow-ups if you'd rather merge the guardrail now and refine it.

proffesor-for-testing:
Update — **`v0.7.1` is now published and is `isLatest` on `meta-proxy-dist`**, so the blocking point above is resolved and this PR just needs its pin retargeted.

`cognitum-one/meta-proxy` v0.7.1 ships the glibc fix (#65). Verified against the **mirrored** artifacts, not a local build:

| image | glibc | aarch64 | x86_64 |
|---|---|---|---|
| `mcr.microsoft.com/devcontainers/base:bookworm` | 2.36 | ✅ | ✅ |
| `ubuntu:22.04` | 2.35 | ✅ | ✅ |
| `debian:buster-slim` | 2.28 | ✅ | ✅ |
| `rockylinux:9-minimal` | 2.34 | ✅ | ✅ |

Signature and all five checksums re-verified after download from the mirror:
```
release_signature=valid public_key_sha256=a0fec4760b81b2819318c17707b02846b88025043a9bccd16dfcdbe37012ddbd
```
The signing key is **unchanged**, so `META_PROXY_SIGNING_PUBKEY_PEM` on this branch is still correct — no re-pin needed.

**One-line change needed here:** `META_PROXY_VERSION = '0.7.0'` → `'0.7.1'`. As written this PR would pin the last release that still carries the `GLIBC_2.39` defect, which is the specific failure the reporter in #149 hit. Say the word and I'll push it to `fix/149-proxy-pin-drift`, or grab it yourself — either way the rest of the PR is good to go.

Also relevant to your watcher: both lookup styles now resolve correctly against the mirror —
```
gh release list --repo cognitum-one/meta-proxy-dist --limit 1  -> v0.7.1 (isLatest=true)
gh api repos/cognitum-one/meta-proxy-dist/releases/latest      -> v0.7.1
```
so it will correctly flag `0.7.0` as drift if merged before the retarget.

Two other things landed on the meta-proxy side that reduce how often this pin can silently rot:
- **Releases now auto-mirror to `meta-proxy-dist`** (cognitum-one/meta-proxy#67) with verify-before-publish and a re-download-and-re-verify afterwards, so "released" and "installable" can't diverge the way they did between 07-17 and 07-27.
- `docs/RELEASING.md` gained the **Downstream** section that both `proxy-pin-drift` workflows link to and which did not previously exist — it now names this constant and this file explicitly as a post-release step.

proffesor-for-testing:
Pushed the retarget to this branch as [`27b0600`](https://github.com/ruvnet/metaharness/commit/27b0600) — one line, `META_PROXY_VERSION` `0.7.0` → `0.7.1`. Nothing else touched; the watcher is untouched and the signing PEM needs no change.

Verified against the **mirrored** artifacts rather than a local build:

- Every URL this constant generates returns `200` on `meta-proxy-dist`:
  ```
  OK 200 darwin-arm64    meta-proxy-0.7.1-aarch64-apple-darwin.tar.gz
  OK 200 darwin-x64      meta-proxy-0.7.1-x86_64-apple-darwin.tar.gz
  OK 200 linux-arm64     meta-proxy-0.7.1-aarch64-unknown-linux-gnu.tar.gz
  OK 200 linux-x64       meta-proxy-0.7.1-x86_64-unknown-linux-gnu.tar.gz
  OK 200 win32-x64       meta-proxy-0.7.1-x86_64-pc-windows-msvc.zip
  OK 200 SHA256SUMS / SHA256SUMS.sig
  ```
- Signature + all five checksums re-verified after downloading from the mirror; key fingerprint `a0fec476…12ddbd` **unchanged**.
- Published binaries start on bookworm / ubuntu:22.04 / buster / rocky9, both arches (8/8).

`__tests__/meta-proxy.test.ts` derives its expected asset name from the constant, so it tracks the bump with no edit.

---

Two things I ran into locally that are **pre-existing on this branch, not from my commit** — flagging rather than fixing, since they're yours to call:

**1. `node_modules` is a tracked symlink to an absolute path.**
```
$ git ls-files -s node_modules
120000 4c37106… 0  node_modules  ->  /home/ruvultra/projects/agent-harness-generator/node_modules
```
It's committed as a symlink (mode `120000`) to a path that only exists on one machine, so a fresh clone anywhere else gets a dangling `node_modules` and `npm ci`/`npm install` fight it (`npm warn reify Removing non-directory .../node_modules`). CI survives only because `npm ci || npm install` re-creates it. Probably wants to be untracked + gitignored.

**2. `npm ci` fails on this branch** — lockfile out of sync, independent of my change:
```
npm error `npm ci` can only install packages when your package.json and package-lock.json are in sync.
npm error Missing: @metaharness/kernel-darwin-arm64@ from lock file
npm error Missing: @metaharness/kernel-linux-x64-gnu@ from lock file  (+2 more)
```
The `|| npm install` fallback in ci.yml masks this, so the lockfile is effectively unenforced — a `npm ci`-only consumer would break.

I confirmed both are baseline by checking against the unmodified branch (`tsc --noEmit` reports the same 202 errors with and without my change), and I reverted the incidental `package-lock.json` and `packages/darwin-mode/dist/cli.js` mode changes my install attempt caused, so this commit is genuinely one line.

The failing `audit-deps-aggregate` check is also pre-existing — 4 high advisories in `apps/web-ui` (`@huggingface/transformers`, `adm-zip`, `onnxruntime-node`, `sharp`), failing on `main` since at least 07-17.

Good to merge from my side once you're happy.

proffesor-for-testing:
CI settled: **24 pass / 1 fail**, and the one failure is the pre-existing `audit-deps-aggregate` (4 high advisories in `apps/web-ui` — `@huggingface/transformers`, `adm-zip`, `onnxruntime-node`, `sharp`), red on `main` since at least 07-17 and unrelated to this branch.

**On the `Node 22 / windows-latest` failure I reported earlier — it was a flake, and I confirmed it rather than assuming.** Re-ran the job with zero code change: `completed/success`. For the record, the evidence beforehand pointed the same way:

- the 12 failures were in `generated-templates` (×5), `host-config` (×2), `mcp-scan`, `secrets`, `validate` (×3) — **zero** reference `meta-proxy`, `META_PROXY` or `proxy` in any of those files
- `__tests__/meta-proxy.test.ts`, the only test that consumes the constant I changed, **passed**
- `Node 22 / ubuntu`, `Node 22 / macOS` and `Node 20 / windows` all passed on the same commit — a version-string change would fail identically everywhere
- the failing tests were the long, filesystem-heavy ones (16–57s on Windows), which is the usual shape of a scaffolding race

Worth knowing independently of this PR: those Windows scaffold tests can fail as a group without a code change, and nothing currently distinguishes that from a real regression. If it recurs, it's probably worth a retry or a timeout bump on the `generated-templates` / `host-config` suites rather than another manual re-run.

So from my side this branch is ready: pin at `0.7.1`, watcher ported, every asset URL verified live against the mirror.

Issue #73, opened 2026-07-03
metaharness CLI bin.js discards main()'s exit code — all commands exit 0


ruvnet:
Found while proofing PR #72: `packages/create-agent-harness/dist/bin.js` (and its source) discards the return code of `main()`, so **every** `metaharness` subcommand exits 0 regardless of failure — a failed command looks green in CI/scripts.

PR #72 scope-fixed this for the new `learn` command only (it sets `process.exitCode` itself rather than relying on the return value). The repo-wide fix is a one-liner in bin.js (`process.exitCode = await main(...)` or equivalent) but changes observable behavior for ALL existing subcommands — some callers may rely on the always-0 today, so it deserves its own change + a quick audit of what `main()` actually returns per command path.

**Acceptance:** `metaharness <cmd>` with a failing sub-dispatch exits non-zero; existing passing flows still exit 0; a regression test pins it.

PR #165, opened 2026-08-03
chore(proxy): pin META_PROXY_VERSION 0.7.3, release metaharness 0.4.5

proffesor-for-testing:
Ready for publish — **this PR does not publish anything**; it stages the tree so `metaharness` can go to npm when you're back at a keyboard.

## What this unblocks

`cognitum-one/meta-proxy` **v0.7.3** is released and mirrored to `meta-proxy-dist`. `metaharness proxy install` downloads the *pinned* version, so until this lands `npx metaharness` keeps installing **0.7.1** and none of these reach a user:

| Fix | What a user gets |
|---|---|
| #75 | Local plane translates Anthropic Messages → OpenAI, so Claude Code can drive a local model |
| #73 | Out-of-credits `402` carries account, tier, cap and reset date instead of a bare refusal |
| #85 | `--help` / `--version` honoured from any argv position |
| #79 | VS Code console prefers the managed daemon over the legacy metaharness path |
| #81 | VS Code console offers a way out when the daemon is not running |

## Why 0.4.5 and not 0.4.4

Tag `v0.4.4` exists in this repo but was **never published** — npm latest is `0.4.3`, because `publish.yml` has not succeeded in any of its 8 runs since 2026-06-14 (`google-github-actions/auth` → `unauthorized_client: The given credential is rejected by the attribute condition`, WIF provider in `ruv-dev`, rejecting both tag refs and `main`). That tag also points at a tree pinning meta-proxy 0.7.2. Publishing a *different* tree under `v0.4.4` would make the tag lie about what shipped, so this takes a fresh 0.4.5.

Happy to redo it as 0.4.4 if you'd rather — say the word and I'll rebase.

## Also fixes pre-existing drift

`package-lock.json` recorded the workspace at `0.4.3`; it was never updated when `package.json` moved to `0.4.4`. Corrected to `0.4.5` **surgically** — running `npm install --package-lock-only` additionally filled in unrelated optional-dependency entries for `kernel-js`, which don't belong in a release change. The lockfile diff here is one line.

## Verified against the real artifact, not just the constant

```
npm run build            (workspace root, then this package)   clean
npm test                 36 files, 410 tests                   pass

HOME=<temp> node dist/bin.js proxy install --yes
  → Installed verified Meta-Proxy v0.7.3
HOME=<temp> node dist/bin.js proxy status
  → installed: yes   version: 0.7.3

# the binary carries the FIX, not just the number (#85):
meta-proxy --port 9999 --help   → prints usage, exit 0
```

"Installed verified" is the meaningful part: the installer fetched from `meta-proxy-dist` and passed **both** the SHA-256 check and the Ed25519 signature check against the public key pinned in this repo. I also verified that independently and unauthenticated from the public download URL — signature `VALID`, tarball sha256 matches `SHA256SUMS`, extracted binary reports `meta-proxy 0.7.3`.

The install ran under a temporary `HOME` so it could not disturb a running daemon. For contrast, the daemon actually installed from npm today still reports `meta-proxy 0.7.1` — that gap is exactly what publishing this closes.

## What's left for you

1. Merge this.
2. Tag `v0.4.5`.
3. `npm publish` by hand — `publish.yml` will fail again at GCP auth, and only `ruvnet` has npm write on `metaharness` (`npm access list collaborators metaharness` → `ruvnet: read-write` only).

If you'd rather fix the automation than keep publishing by hand, the block is the WIF attribute condition on `projects/875130704813/.../workloadIdentityPools/github-pool/providers/github` for `metaharness-publish@ruv-dev.iam.gserviceaccount.com` — it rejects this repo's tag refs *and* `main`, which is why the workflow has never once succeeded.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5013,7 +5013,7 @@
     },
     "packages/create-agent-harness": {
       "name": "metaharness",
-      "version": "0.4.4",
+      "version": "0.4.5",
       "license": "MIT",
       "dependencies": {
         "@metaharness/darwin": "^0.8.0",
--- a/packages/create-agent-harness/package.json
+++ b/packages/create-agent-harness/package.json
@@ -1,6 +1,6 @@
 {
   "name": "metaharness",
-  "version": "0.4.4",
+  "version": "0.4.5",
   "description": "MetaHarness \u2014 mint a custom AI agent harness from any repo. Browser Studio + `npx metaharness` CLI. Runs on Claude Code, Codex, pi.dev, Hermes, OpenClaw, RVM, Prime Agent.",
   "homepage": "https://github.com/ruvnet/agent-harness-generator",
   "repository": {
--- a/packages/create-agent-harness/src/meta-proxy.ts
+++ b/packages/create-agent-harness/src/meta-proxy.ts
@@ -26,7 +26,7 @@ import {
 import { homedir } from 'node:os';
 import { basename, join } from 'node:path';
 
-export const META_PROXY_VERSION = '0.7.2';
+export const META_PROXY_VERSION = '0.7.3';
 export const META_PROXY_RELEASE_BASE = 'https://github.com/cognitum-one/meta-proxy-dist/releases/download';
 
 /** The release signing key, pinned in the client rather than fetched from GitHub. */


proffesor-for-testing:
### On the red `audit-deps-aggregate` check — it is not from this PR

```
[audit-deps] PASS: npm(workspace)   — 0 advisories at-or-above high
[audit-deps] PASS: cargo            — 0 Rust advisories at-or-above high
[audit-deps] FAIL: npm(apps/web-ui) — 4 advisories at high+:
             @huggingface/transformers(high), adm-zip(high),
             onnxruntime-node(high), sharp(high)
```

The `Security` workflow is already failing on `main` — runs 30800417065 (09:12) and 30810926791 (11:47) today, both before this branch existed. And this PR's entire diff is three files:

```
package-lock.json
packages/create-agent-harness/package.json
packages/create-agent-harness/src/meta-proxy.ts
```

Nothing under `apps/web-ui`. The workspace audit that *does* cover the package this PR changes passes with 0 advisories.

I've deliberately not folded a fix for those four into a release pin bump — they are a separate piece of work in a different app, and bundling them would make this harder to review and to revert. Happy to open a separate PR for `apps/web-ui` if you want it picked up.

For contrast, a genuinely-in-scope advisory *did* surface while cutting meta-proxy v0.7.3 and was fixed there rather than deferred: `brace-expansion@5.0.8` / GHSA-rgw5-rvv9-x895 turned three jobs red in `cognitum-one/meta-proxy`, and landed as a 3-line lockfile refresh to 5.0.9 (minimatch@10 already asks for `^5.0.5`, so no override was needed — an override there is precisely what once broke `glob` at runtime while the scan went green).

proffesor-for-testing:
Status update — **this PR needs no change**. It already pins `META_PROXY_VERSION = '0.7.3'`, and 0.7.3 is still the latest published meta-proxy daemon. It is correct as-is and blocked only on merge + `npm publish`.

Two things changed on the meta-proxy side today that are worth knowing before you merge:

**1. There is no newer daemon to re-pin to.** The only commit since `v0.7.3` touched a single file (`vscode-extension/package-lock.json`, a dev-only advisory refresh). `git diff v0.7.3..main -- src/ Cargo.toml Cargo.lock build.rs` is empty, so a `v0.7.4` would have been a byte-identical binary and a pointless re-pin here. Deliberately not cut.

**2. The VS Code extension has been released: [`vscode-v0.7.3`](https://github.com/cognitum-one/meta-proxy/releases/tag/vscode-v0.7.3)** — the first extension release since `vscode-v0.7.0`, shipping three unreleased versions (the stale-binary path-ordering fix, the console failure-classification fix, and the advisory refresh). Signed VSIX + `SHA256SUMS` + `SHA256SUMS.sig`; I verified the checksum and the Ed25519 signature from the public release, not from CI.

That release raised the extension's `MINIMUM_DAEMON_VERSION` from 0.7.1 to **0.7.3**, which the release workflow requires. **That is why publishing this PR matters more than it did yesterday.**

While npm's latest is `metaharness@0.4.3` (pinning 0.7.1), a user running the extension gets *"Meta Proxy 0.7.3+ required"* and has to run an explicit upgrade command. Measured against the real 0.4.3:

```
metaharness proxy install --version 0.7.3          exit 2, installs nothing
metaharness proxy install --yes                    installs 0.7.1  ← still rejected
metaharness proxy install --version 0.7.3 --yes    installs 0.7.3  ← works
```

The extension previously printed only the first form, which was a dead end — fixed in the same release, so both of its install instructions now carry `--version <v> --yes`. So it is one extra command, not breakage.

**Publishing 0.4.5 removes even that**: the bare `install` then delivers 0.7.3 and the floor is satisfied by default. No re-pin needed — 0.7.3 is already the right value.

Also flagging the deliberate ordering, since it constrains a step on our side: we have **not** dispatched the VS Code Marketplace publish, and will not until 0.4.5 is on npm. The tag only stages a signed artifact; nobody auto-updates into the new floor. That ordering is recorded in the code so it does not get lost.

Nothing needed from you beyond merge + `npm publish` — noting again that `publish.yml` has never succeeded (WIF `unauthorized_client`), so this is a manual publish.

zainnabcheema6-maker (approved):

zainnabcheema6-maker:
looks goods overall

zainnabcheema6-maker:
For what  purpose is used

proffesor-for-testing:
Rebased onto `main` (`a9fe411`) — the branch was `CONFLICTING`, it is now `MERGEABLE` / `CLEAN` with **25/25 checks green**, including `audit-deps-aggregate`, which was the one red check before and which `main` has since fixed.

**What the conflict actually was.** Only the version line, in `package.json` and `package-lock.json`. `meta-proxy.ts` never conflicted. The original branch also un-escaped the em dash in `"description"`, which is what turned a one-line version bump into a collision with #176/#177's description edit. That cosmetic change is dropped — `main`'s description, the `prime-agent` keyword and the `@metaharness/turn-credit` dependency are kept exactly as they are on `main`. The result is a **3-line diff**.

**Two things in the original description are now out of date**, since 0.4.4 was published while this sat:

- it said `npx metaharness` installs meta-proxy **0.7.1** — it installs **0.7.2** today
- it justified 0.4.5 on the grounds that 0.4.4 was never published. 0.4.4 *is* published now, and it pins 0.7.2, so it cannot carry this change. 0.4.5 is still the right version, for a different reason.
- the lockfile drift it also fixed (workspace recorded at 0.4.3) was corrected on `main` independently; this now just moves 0.4.4 → 0.4.5.

The commit message has been rewritten so none of those stale claims ship with the release commit.

**Re-verified after the rebase, against the artifact rather than the constant:**

```
npm run build                     clean, all 4 phases
npm test                          2243 tests, 37 workspace suites, exit 0
HOME=<temp> proxy install --yes → "Installed verified Meta-Proxy v0.7.3"
HOME=<temp> proxy status        → version: 0.7.3
meta-proxy --port 9999 --help   → usage, exit 0     ← that IS #85
```

"Installed verified" means it fetched from `meta-proxy-dist` and passed **both** the SHA-256 check and the Ed25519 signature check against the pinned public key. It ran under a temporary `HOME` so it could not disturb any daemon already installed on this machine.

**Unrelated finding, worth a separate issue:** `npm ci` fails on this repo, and it fails identically on pristine `main` — `Missing: @metaharness/kernel-darwin-arm64@ … kernel-win32-x64-msvc@ from lock file`. The five `kernel-js` optional platform deps are not recorded in `package-lock.json`, so the repo cannot be clean-installed. Not introduced here and deliberately not fixed here, since it does not belong in a release pin bump.

**Still owed and only you can do it:** merge this, tag `v0.4.5`, and `npm publish` by hand. `publish.yml` has never succeeded — 8 runs since 2026-06-14, all failing at `google-github-actions/auth → unauthorized_client: The given credential is rejected by the attribute condition`, because the WIF provider in `ruv-dev` rejects both tag refs and `main`. `npm access list collaborators metaharness` shows only `ruvnet: read-write`.

ruvnet:
Superseded by #175, now merged (87fbb4d) — it carries the same 0.4.5 bump but pins the newer signed Meta-Proxy v0.7.4 (published + mirrored to meta-proxy-dist today via cognitum-one/meta-proxy#89 and tag v0.7.4) plus the full service-lifecycle contract. Verified end-to-end: `proxy install --yes` under a temp HOME reports "Installed verified Meta-Proxy v0.7.4" (SHA-256 + Ed25519 against the pinned key). One correction to this PR's rationale for the record: 0.4.4 *was* subsequently published to npm (2026-08-09, via #169), so npm latest was 0.4.4, not 0.4.3 — the 0.4.5 target was right anyway. Thanks — the lockfile-drift fix and the verification recipe here carried straight into the release.

PR #167, opened 2026-08-04
fix(darwin): preserve mock sandbox during bench evaluation

### LINKED ISSUE #166: Darwin --sandbox mock --bench executes real benchmark commands and exceeds wrapper timeout
## Environment

- `@metaharness/darwin@0.8.0` through Ruflo MetaHarness evolve wrapper
- macOS arm64, Node 22
- verified bench: 30 tasks, hash accepted
- isolated repository copies created from the same immutable Git revision

## Reproduction

Four independent runs used deterministic mutation, mock sandbox, 3 generations, 5 children, concurrency 4, and distinct seeds/selection strategies (`quality-diversity`, `pareto`, `behavioral-diversity`, `clade`). Each generated the baseline plus five generation-1 variant directories and only `runs/baseline.json`; none produced child run records or a winner. Processes remained alive beyond the wrapper's 180000 ms timeout and required targeted termination.

A reduced run also reproduced:

- 1 generation
- 2 children
- concurrency 1
- deterministic mutator
- mock sandbox
- pareto selection
- 60000 ms wrapper timeout

The wrapper returned `degraded: true`, `reason: metaharness-darwin-timeout` after roughly 57 seconds. No candidate was promoted.

## Expected

Mock-sandbox evolution should either complete and emit child run records/winner, or terminate all child processes at the configured deadline with a diagnostic naming the blocked phase/task.

## Safety note

The corpus contains a hidden command that intentionally exits non-zero while human adjudication is pending. In mock mode this should not hang; if a non-zero hidden command is the trigger, it should be recorded as a failed task/variant.

proffesor-for-testing:
## Summary

Fixes #166 by preserving `sandboxMode: 'mock'` during benchmark-suite evaluation.

Normal Darwin evaluation already routes mock mode through `runVariantTasksMock`, but the optional bench phase always called `evaluateChildAgainstParent`, which executes each task's public, hidden, and regression shell commands. A 30-task suite therefore ran up to 90 real commands per parent/child comparison despite `--sandbox mock`.

This change:

- maps hash-pinned benchmark tasks into the deterministic mock manifold
- retains benchmark identity, commit provenance, scoring, statistical comparison, and promotion gates
- never executes benchmark shell commands in mock mode
- keeps real/agent behavior unchanged
- updates the shipped `dist` and ADR-102
- adds unit and marker-file end-to-end regressions

## Verification

- `npm run build -w @metaharness/darwin`
- `npm run lint -w @metaharness/darwin`
- `npm test -w @metaharness/darwin -- --run`
  - 67 files passed
  - 593 tests passed, 14 skipped
- focused regression: 12/12 passed
- actual Cognitum 30-case bench, 1 generation / 2 children / concurrency 1 / deterministic mutator / mock sandbox:
  - completed in 2.2 seconds
  - emitted both child `.bench.json` decisions and winner artifacts
  - no benchmark commands executed

## Repository note

Current upstream `main` has a pre-existing package-lock mismatch for optional `@metaharness/kernel-*` packages, so `npm ci` fails before tests. Dependencies were installed with `npm install --package-lock=false --ignore-scripts`; the lockfile is unchanged.


zainnabcheema6-maker (approved):

zainnabcheema6-maker:
looks good overall

PR #175, opened 2026-08-10
fix(proxy): harden managed service lifecycle

proffesor-for-testing:
## Summary
- make service-manager state authoritative over definition-file presence
- compensate partial enablement and preserve evidence on failed disable/uninstall
- add bounded log tail and explicit-consent clean uninstall that removes only owned files
- pin the compatible Meta-Proxy daemon at 0.7.4 and bump the CLI to 0.4.5
- add a collision-free real macOS LaunchAgent acceptance test

## Adversarial evidence
- focused lifecycle, CLI, workflow, and Windows PATH-timeout tests: 46 passed; real LaunchAgent acceptance passed separately
- real isolated label proved RunAtLoad, SIGKILL restart with a new PID, managed stop staying stopped beyond the throttle interval, explicit restart, disable, and cleanup
- clean `npm ci`, full ordered build, and reproducible lockfile passed
- full workspace test passed
- package lint passed
- `npm audit --omit=dev --audit-level=high`: 0 vulnerabilities
- `git diff --check`

The real test uses a unique `one.cognitum.meta-proxy.qe.*` label and temporary home. A `finally` bootout plus recursive removal guarantees cleanup; it never addresses the normal user label.

## Release dependency / rollback
Blocked on cognitum-one/meta-proxy#89 and publication of its signed v0.7.4 artifacts. Do not merge or publish this package while that asset is absent.

This clean successor supersedes #165 without rewriting its user-owned branch. Rollback is MetaHarness 0.4.4 pinned to the previous signed v0.7.3 daemon; no user credential or routing-state migration is performed.

Relates to cognitum-one/meta-proxy#82. Keep the issue open until the signed daemon and paired package pass native acceptance.

### Gate 3 remediation
- separates manager load, active/PID, and login-enable state; active disabled systemd units remain stoppable
- Linux reload failure restores the exact unit; Windows ends and confirms stopped before delete and preserves evidence on uncertainty
- required aggregate macOS CI job runs the real isolated LaunchAgent lifecycle

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -177,6 +177,27 @@ jobs:
         shell: bash
         run: node examples/vertical-tour/vertical-tour.mjs
 
+  # Real user LaunchAgent contract. This must stay a required, non-skipped job:
+  # unit render tests cannot prove launchd restart and bootout behavior.
+  meta-proxy-launchagent:
+    name: Meta-Proxy / native macOS lifecycle
+    runs-on: macos-latest
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262
+      - uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020
+        with:
+          node-version: '20'
+      - name: Install exact dependencies
+        run: npm ci
+      - name: Build MetaHarness CLI
+        run: npm run -w packages/create-agent-harness build
+      - name: Exercise real isolated LaunchAgent lifecycle
+        working-directory: packages/create-agent-harness
+        env:
+          RUN_REAL_LAUNCHD_ACCEPTANCE: '1'
+        run: npx vitest run __tests__/meta-proxy-launchagent.real.test.ts
+
   # ─────────────────────────────────────────────────────────────────────
   # Per-platform npm pack + install smoke for every published package
   # ─────────────────────────────────────────────────────────────────────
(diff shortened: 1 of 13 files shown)


ruvnet:
Reviewed at a1e6e53. Three findings, fixed in fa848f1:

1. **Foreign-service bootout on darwin/linux** — `enableMetaProxyService` only guarded win32 against a label-matching job with no owned definition. On macOS, a job registered by another installer made `before.enabledAtLogin` false (it's defined as `definitionPresent` there), so the line-439 guard never fired; a failed `launchctl bootstrap` then "compensated" with `bootout`, stopping the user's running proxy. Linux had the same shape via a shadowing system unit. The ownership guard now applies on every platform.
2. **`disable`/`uninstall --yes` unregistered by label alone on darwin/linux** — the win32-only refusal ("registered without the owned definition") is now cross-platform.
3. **Windows uninstall EBUSY** — `rmSync(binary)` on a still-locked exe threw mid-loop, leaving the version/pid/log/task files behind and surfacing a raw exception (this PR's own scheduled-task acceptance test needed a 5s retry loop for exactly this). Uninstall now retries locked-file codes briefly and reports unremovable paths with a re-run hint.

5 new regression tests; package build + 452 tests green.

Remaining blocker is unchanged and external: signed meta-proxy v0.7.4 artifacts don't exist in meta-proxy-dist yet (latest is v0.7.3). Working that via cognitum-one/meta-proxy#89 next. Note `proxy-pin-drift.yml` will file an inverted "pinned 0.7.4 behind latest 0.7.3" alert on Mondays until the release lands.

PR #179, opened 2026-08-13
docs: ADR-251 — the MetaHarness Dream Machine (nightly cloud-scheduled evolution routine)

ruvnet:
## What

Ports the Ruflo Dream Cycle v3 architecture (gist `889ffa92dab49d508e70b123c940e1b9`) to this repository:

- **`docs/adrs/ADR-251-nightly-dream-machine.md`** — the decision record (Proposed), following the repo ADR shape incl. Test Contract, plus the INDEX.md row.
- **`docs/dream-cycle/PROMPT.md`** — version-controlled mirror of the routine prompt (the cloud scheduler copy is authoritative; the header requires both to be updated together).

## The routine

"MetaHarness Nightly Dream Cycle", cron `0 8 * * *` UTC (staggered after Ruflo's 06:00), fresh isolated cloud checkout, one nightly pipeline:

ledger → research → frozen hypothesis → concrete candidate → baseline → evaluation → adversarial critique + reward-hack check → bounded Darwin → evidence retention → witness → gist + issue + **draft** PR + one ledger row.

Key repo-specific deltas: first-class `LLM_EVAL=blocked` and JS-only-build degraded modes, ADR-250 proof-ladder rung on every quantitative claim, surface rotation over generator-genome / flywheel-promotion / darwin-evolution / security-adversarial / host-adapters.

The routine never merges, never self-promotes, never weakens a test. `docs/dream-cycle/LEDGER.md` is created by the first run.


ruvnet:
Reviewed and conflict-free against main, but it had **two real defects**. Both fixed in `b3961ddd`; both were caught by `__tests__/adr-index.test.ts`, which **never runs in CI** — it lives in the root suite, which `npm run -ws` excludes and which #195 has now formally allowlisted as unreached. So neither would have surfaced on merge.

**1. ADR number collision.** `ADR-251` is already taken on `main` by `ADR-251-governed-autonomous-variation-runtime.md` (`@metaharness/avo`, Accepted and shipped). This PR added a **second** ADR-251, and `INDEX.md` ended up with two rows carrying the same number. It is not merely cosmetic: `ADR-252` (LatentMesh) already cites *"edge authority mapped onto ADR-251 capability policy"*, which means the AVO one — a duplicate makes that existing cross-reference ambiguous.

Renumbered to **ADR-273** (272, the darwin numeric genome kind, is the current high-water mark). The AVO ADR keeps 251, so every existing reference still resolves to what it always meant.

**2. Non-conforming headings.** The ADR used numbered sections — `## 1. Context`, `## 2. Decision`, `## 3. Consequences`. I measured the repo before changing anything: **235 of 236 ADRs use the unnumbered form**, and this file was the sole exception. The test encodes the real convention, so the file was wrong, not the test. Numeric prefixes stripped.

**Also corrected:** the INDEX row claimed *"Proposed — routine created; first live run pending"*, while `LEDGER.md` records **10 completed runs through 2026-08-29** (and dream PRs run to #258 on 09-01). Updated to match the ledger.

Checked and left alone: `docs/dream-cycle/PROMPT.md` is already **v2** (Ruflo Dream Cycle v3.1), matching the `Updated: 2026-08-14` note in the ADR and the v2 behaviours visible in the ledgered runs — so the mirror is current, not drifted.

---

**Separate finding, not fixed here.** `ADR-253-arc-agi-3-chatgpt-harness.md` has **no Consequences section** and fails the same test on `main` today. It has no section that could be renamed into one (`Context` → `Decision` → domain sections → `References`), so writing that analysis is the author's call, not a mechanical fix. Filing it separately along with the more important point: this is the second defect class in a row that hid behind the root suite never executing.

Issue #20, opened 2026-06-17
kernel npm package ships no wasm/native artifacts — backend always falls to js


ruvnet:
## Summary
`@metaharness/kernel@0.1.0` (npm) ships **only the pure-JS floor backend**. The native (NAPI) and WASM artifacts are produced by separate CI jobs and are **not in the published tarball**, so a clean `npm install` can never reach the `native` or `wasm` backends — even though the generated `doctor` advertises `native|wasm|js`.

## Repro
```bash
npx metaharness demo --template minimal --host claude-code --target /tmp/demo
cd /tmp/demo && npm install && node bin/cli.js doctor
# => "... js backend" on every machine; METAHARNESS_KERNEL_BACKEND=wasm has no effect
```
`loadKernel()` resolves native → wasm → js; with no artifacts present it always lands on `js`.

## Impact
Every generated harness silently runs JS-only. "WASM kernel" is effectively unavailable to end users via npm.

## Suggested fix
- Publish the prebuilt `pkg/ruflo_kernel_wasm.*` inside `@metaharness/kernel` (it's small), and/or publish the per-platform NAPI optional-dependency packages (`@metaharness/kernel-<triple>`), so `npm install` can resolve a non-JS backend.
- If that's intentional for the beta, have `doctor` say so explicitly ("native/wasm artifacts not published; using js floor").

PR #189, opened 2026-08-14
fix: root build:wasm delegates to the kernel's authoritative nodejs-target script

### LINKED ISSUE #173: README dev quickstart fails on a clean clone: root `npm run build:wasm` builds --target bundler into the loader's out-dir, so `npm test` fails the #20 regression guard
## Summary

The README's developer quickstart fails on a clean clone. `npm run build:wasm` **from the repo root** builds the `bundler` target into `packages/kernel-js/pkg`, which the runtime loader cannot load — so the next command in the same block, `npm test`, fails on the repo's own regression test for #20.

Two scripts share the name `build:wasm` and the same `--out-dir`, with **opposite targets**:

| script | target | result |
|---|---|---|
| `package.json:27` (root) | `--target bundler` | `pkg/package.json` → `"type": "module"` — loader throws |
| `packages/kernel-js/package.json:44` | `--target nodejs` (+ `rm -f pkg/.gitignore`, forces `"type":"commonjs"`) | loads correctly |

`.github/workflows/publish.yml:99-103` already identifies the second as authoritative:

> ``A `bundler`-target artifact is ESM that needs a bundler + an explicit init() call, so loadWasm() cannot load it and the kernel silently falls back to the JS floor — the exact #20 symptom. `npm run build:wasm` is the single source of truth (correct target + the .gitignore/package.json fixups).``

...and runs it with `working-directory: packages/kernel-js`. The root script is the one a reader reaches from the quickstart, and it reintroduces precisely the symptom that comment warns about.

## Reproduction — the README's own block, verbatim

```bash
git clone https://github.com/ruvnet/metaharness && cd metaharness
cargo test --workspace                                  # ok — 166 passed
cargo clippy --workspace --all-targets -- -D warnings   # ok — 0 warnings
npm install                                             # ok
npm run build:wasm                                      # exits 0, WRONG TARGET
npm test                                                # FAILS
node scripts/healthcheck.mjs                            # HEALTHY (8/8)
```

Failure:

```
FAIL __tests__/loader.test.ts > GH #20 — publish ships a loadable wasm (nodejs target)
     > a present wasm artifact resolves the wasm backend
Error: kernel backend "wasm" was requested but is unavailable:
       wasm artifact (pkg/ruflo_kernel_wasm.js) not present in @metaharness/kernel
       ("ESM integration proposal for Wasm" is not supported currently...)

Test Files  1 failed | 6 passed (7)
     Tests  1 failed | 58 passed (59)
npm error workspace @metaharness/kernel@0.1.3
```

Everything else is green: ~2,131 npm tests across 37 workspaces pass, only this one fails.

## Root cause confirmed by rebuilding

Not inferred — switched only the build command and re-ran the same test:

```bash
cd packages/kernel-js
# pkg/package.json "type" BEFORE: module      (root bundler build)
npm run build:wasm
# pkg/package.json "type" AFTER:  commonjs    (nodejs build)
npx vitest run __tests__/loader.test.ts
#  Test Files  1 passed (1)
#       Tests  15 passed (15)     <- was 1 failed | 14 passed
```

One variable changed; the failure flips deterministically.

## Why it matters

- **The quickstart is the first thing a contributor runs**, and it ends red on an untouched checkout. The natural reading is "the repo is broken", not "you ran the wrong one of two identically-named scripts".
- The failing test is the guard for #20. A contributor who sees it fail after following the docs may reasonably `--skip` it or assume it is flaky — which removes the regression guard for the exact bug it protects.
- Root `build:wasm` writes into the same `--out-dir` as the authoritative one, so it silently **overwrites** a correct artifact with an unloadable one.

## Suggested fix

Any of, cheapest first:

1. **Make the root script delegate**: `"build:wasm": "npm run build:wasm -w @metaharness/kernel"` — one line, removes the divergence, keeps the quickstart working as written.
2. If the bundler artifact is genuinely wanted for a separate consumer, give it a distinct name **and a distinct `--out-dir`** (e.g. `build:wasm:bundler` → `pkg-bundler/`) so it cannot clobber the loadable one.
3. Failing either, correct the README block to `npm run build:wasm -w @metaharness/kernel`.

## Environment

- `ruvnet/metaharness@6db0923` (clean `--depth 1` clone)
- rustc 1.88.0, wasm-pack 0.15.0, node v22.23.0, npm 10.9.8, darwin arm64

## Related

- #20 (closed) — the regression this test guards; the failure message is that symptom.
- #172 — the published `@metaharness/kernel` still ships no `pkg/`; this issue is about the *local build*, which does produce a valid 174 KB artifact when the right script is used.

## Minor, same run

`scripts/healthcheck.mjs` prints `Result: HEALTHY (8/8 pass)` when one of the eight is `SKIP pages` — 7 passed, 1 skipped. Counting a skip as a pass makes the summary line unable to distinguish "checked and fine" from "not checked".

ruvnet:
## Summary

Fixes #173. The root `build:wasm` script built `--target bundler` into `packages/kernel-js/pkg` — an ESM artifact `loadWasm()` cannot load — so the README quickstart's `npm run build:wasm && npm test` failed the GH #20 regression guard on every clean clone. `publish.yml` already treats `packages/kernel-js`'s script (nodejs target + `.gitignore`/`package.json` fixups) as the single source of truth; the root script now delegates to it instead of shadowing it with the opposite target.

## Verification

- Clean `npm install && npm run build:wasm` from the repo root now emits `pkg/` with `"type": "commonjs"` and no stray `.gitignore`
- `packages/kernel-js` `loader.test.ts` — including the #20 guard ("publish ships a loadable wasm (nodejs target)") — passes: 15/15

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "build": "node scripts/build-ordered.mjs",
     "test": "npm run -ws --if-present test",
     "lint": "npm run -ws --if-present lint",
-    "build:wasm": "wasm-pack build crates/kernel-wasm --target bundler --out-dir ../../packages/kernel-js/pkg --release",
+    "build:wasm": "npm run build:wasm --workspace @metaharness/kernel",
     "build:napi": "cd crates/kernel-napi && napi build --platform --release --output-dir ../../packages/kernel-js/native",
     "test:rust": "cargo test --workspace",
     "fmt:rust": "cargo fmt --all",


vidaunited:
Confirmed fixed — I filed #173, and re-ran the README quickstart end-to-end on a clean clone of `main` @ `1f3b870` (post-merge, working tree clean, macOS, node v22.23.0, wasm-pack 0.15.0, cargo 1.97.1).

**The quickstart now passes:**

```
npm install     rc=0
npm run build:wasm    rc=0   → pkg/package.json "type": "commonjs"
npm test        rc=0   → 2301 passed | 0 failed | 59 skipped, 38 result blocks
```

That was the exact pair that failed before, so #173 is closed correctly.

**I also proved the test can still fail**, because a confirmation where the guard is simply inert would be worthless. Three states, one variable — the build command — everything else identical:

| `packages/kernel-js/pkg` | `"type"` | `__tests__/loader.test.ts` |
|---|---|---|
| absent | — | `14 passed \| 1 skipped (15)` |
| built with the **old** root command (`--target bundler --out-dir ../../packages/kernel-js/pkg`) | `module` | **`1 failed \| 14 passed (15)`** |
| built with the **new** root command (`npm run build:wasm`) | `commonjs` | `15 passed (15)` |

The middle row is the original #173 failure reproduced verbatim on today's `main`, same message:

```
FAIL  __tests__/loader.test.ts > GH #20 — publish ships a loadable wasm (nodejs target)
      > a present wasm artifact resolves the wasm backend
Error: kernel backend "wasm" was requested but is unavailable: wasm artifact
       (pkg/ruflo_kernel_wasm.js) not present in @metaharness/kernel
       ("ESM integration proposal for Wasm" is not supported currently...)
```

So the delegation is doing the work, not masking it. I also checked the trap can't come back by another route: six `--target bundler` occurrences survive on `main`, and none can reach the loader's directory — `scripts/preflight.mjs:128` and `.github/workflows/ci.yml:101` both build with **no `--out-dir`**, so their artifact lands in `crates/kernel-wasm/pkg` and CI validates it there. The defect was never the bundler target; it was a bundler artifact written into `packages/kernel-js/pkg`, and nothing does that any more.

---

**One observation you may or may not want to act on** — offered as context for why this survived so long, not as a request to reopen anything.

Note row 1 of the table: with **no** artifact present, the #20 guard **skips** rather than fails. So the guard only bites when an artifact exists and is wrong; it is silent when nothing was built at all. That is a defensible design — you can't assert on an artifact that's optional at test time — but it does mean the bad root script produced a green `npm test` for everyone who never ran `build:wasm`, and the failure only surfaced for someone following the quickstart in order. If you ever want that closed, a check that the artifact is present *when the workspace is being tested for publish* would catch the "never built" case that the current skip lets through.

Thanks for the quick turnaround on the one-line fix — delegating to a single authoritative script is the right shape, and it matches what `publish.yml` already called the source of truth.

vidaunited:
Filed the skip-guard observation from my previous comment as its own issue — #192 — so it isn't carried in this PR thread. It includes a scoped regex that catches the #173 shape while leaving the legitimate bundler builds in `preflight.mjs` and `ci.yml` alone (checked against all four real call sites: one flag, three passes). Nothing needed here; #189 stands as merged.

Issue #199, opened 2026-08-15
[Dream Cycle 2026-08-15] generator-genome: scorePublishReadiness locked python/go out of 'ready' + router,turn-credit scan


ruvnet:
## Tonight's Rotation

```
DATE=2026-08-15
DEEP=generator-genome (scaffolding quality, genome scorers, score/diag/doctor)
SCAN=router,turn-credit
SLOT=0 (DAYINT % 5)
COMMIT=5f1d4b244cff5af0ec9b147a3f57f615b45abc22 (HEAD @ start of run, main, after correcting a stale local branch ref — see Ledger Check)
BRANCH=dream/2026-08-15-generator-genome
```

Build: full JS/TS workspace build green (`npm run build`, 0 errors). No Rust/wasm/NAPI step encountered or needed for tonight's candidate. `OPENROUTER_API_KEY` was present → `LLM_EVAL=available`, but tonight's candidate needed zero model calls (deterministic scorer-correctness fix).

**Environment note, disclosed not hidden**: this container's local `main` branch ref was stale (pinned at `68402755`/metaharness 0.4.5, 11 commits behind). The initial detached-HEAD checkout at session start was already correct (`5f1d4b2`/0.4.7), but running `git checkout main` moved onto the stale local branch. Caught via a discrepancy between two research subagents' reported `git log` output, fixed with `git fetch origin main && git merge --ff-only origin/main` (local `main` had zero unique commits, pure fast-forward, unaffected). Also affected: an early research subagent's "Finding 1" (taskCoverage still buggy) was based on the stale checkout and is **invalid** — `#171`/`#196`/e66e2b2 are in fact already merged and well-tested on real `main`. Not shipped as a finding.

## Ledger Check

`docs/dream-cycle/LEDGER.md` had 2 prior rows (2026-08-13 security-adversarial, 2026-08-14 host-adapters). Both PRs (#181, #188) confirmed **MERGED** via `pull_request_read`/`list_pull_requests`. Prior gist (2026-08-14) scored **10/10** against the rubric: evaluation grade A-equivalent (deterministic repro) +2, 6 competitor rows +2, 3 specific recommendations +2, valid witness (round-trip re-verified) +2, 1495 words (<1500) +1, genuinely novel finding +1. No STEP 1.1 learning-signal thresholds trip (only 2 prior nights, no duplicate direction, no zero-merge streak, no low-score streak, no LLM_EVAL=blocked streak).

## Deep Dive Findings

[...]

PR #200, opened 2026-08-15
dream(generator-genome): scorePublishReadiness locked python/go out of 'ready' (evaluated)

ruvnet:
MetaHarness Dream Cycle, 2026-08-15. Deep surface: `generator-genome` (slot 0). Full detail, receipts, candidate-selection scoring, adversarial-critique writeup, and router/turn-credit scan findings: #199.

## Hypothesis

Given `genome-scorers.ts`'s `scorePublishReadiness(profile, plan)` and `analyze-repo.ts`'s `buildCommands` inference, when the language-credit term is generalized from a `typescript`/`rust` allow-list to "any detected language" and `buildCommands` gains a python (`pip install -e .` / `pip install -r requirements.txt`) and go (`go build ./...`) entry, then a fully mature python or go repo (CI-wired, tested, buildable, default-deny policy) should score `publish_readiness >= 0.75`, matching an equivalently mature typescript repo, subject to: (a) typescript/rust scores do not regress, (b) all 5 `genome-scorers.ts` functions gain direct unit coverage (zero existed before), (c) no existing test weakened, (d) diff stays materially under 300 lines.

## Candidate

`scorePublishReadiness` credited language detection only for `typescript` (+0.3) and `rust` (+0.15); every other language got `+0`, and `analyze-repo.ts` never generated a `buildCommands` entry for python/go, so their `+0.2` buildable bonus was structurally unreachable too. A fully mature python/go repo capped at 0.40 publish-readiness — `genome.ts` requires `>=0.75` for `'ready'`, so no pure-python/go repo could ever reach that verdict. Fixed: generalized the language credit to "any detected language" (`genome-scorers.ts`), added real build-command inference for python/go (`analyze-repo.ts`), added `packages/create-agent-harness/__tests__/genome-scorers.test.ts` — the first dedicated unit suite for `genome-scorers.ts`'s 5 pure functions, 19 tests. 3 files, ~200 lines total (well under the 300-line target).

## Evaluation Receipt

Deterministic reproduction (not sampled):
- **Pre-fix**: 16/18 passed, 2 predicted failures (both on `buildCommands.length > 0`) — non-vacuous, independently re-derived by the adversarial critic via direct revert-and-rerun.
- **Post-fix**: 19/19 passed (18 + 1 added post-critique to close a generalization-coverage gap).
- Full package suite: **477 passed, 2 pre-existing skips**, 0 regressions.
- `tsc --noEmit`: clean. Full repo `npm run build`: clean.

## Baseline Comparison

All pre-existing tests in the package pass unchanged. Rust's publish-readiness credit rises 0.15→0.3 as a side effect of the generalization (no longer arbitrarily lower than typescript's) — independently verified by the critic to cause **no verdict-threshold flip** in either direction: a CI-wired rust repo was already exactly at the 0.75 `'ready'` cutoff pre-fix (0.90 post-fix); without CI, 0.55→0.70, still below 0.75 both times.

## Darwin Lineage

Not run — scorer-correctness fix, not a tunable routing/topology/prompt/tool/tier parameter; no Darwin mutation surface applies (ADR-071 mutation allowlist).

## Flywheel Evidence

Not applicable — `generator-genome` is not a flywheel domain.

## Reward Hack Check

Purely additive diff, no existing test weakened. Independent critic attempted to construct a "cheating" implementation (a 4-language allow-list masquerading as a generalization) that would pass the original test suite — found it would, closed with an added 19th test that constructs a hand-built `RepoProfile` for a language `analyzeFiles()` has never heard of.

## Security Review

No credential/shell/network surface touched, no new dependency. The new `buildCommands` suggestions (`pip install -e .`, `go build ./...`) are advisory-only and never auto-executed (`execution: 'disabled'`, confirmed via `exec`/`spawn` grep). Raising publish-readiness for now-fairly-scored repos is a fairness correction — the independent MCP-risk/default-deny safety terms are untouched.

## Regression Analysis

Zero regression: 477/479 (2 pre-existing skips) unaffected outside the touched files; `tsc` and full build clean.

## ADR

None — this is a scorer correctness fix (parameter/heuristic-scale), not an architectural decision, per the Dream Cycle prompt's own guidance on when an ADR is warranted.

## Research Gist

No gist-creation tool available in this session's toolset — committed instead: [`docs/dream-cycle/2026-08-15-gist.md`](https://github.com/ruvnet/metaharness/blob/dream/2026-08-15-generator-genome/docs/dream-cycle/2026-08-15-gist.md) (`GIST=LOCAL`).

## Issue

#199 (full findings, candidate-selection scoring table, router/turn-credit scan results, competitor review).

## Witness

```
session_commit : 5f1d4b244cff5af0ec9b147a3f57f615b45abc22
report_sha256  : b0ede37fe44574a2eb9de78ad629d244a7403edda258b9b9cdd1cea6c3275178
witness        : b8c113654113145c6180e57a7b3c2f5fedf0d1f1fc6cb005f974f4fc17141a0f
```
Verifier procedure is in the gist's Witness section and was independently self-verified (round-trip reproduced) before this PR was opened.

## Merge Policy

Recommendation: **ACCEPT-WITH-CAVEATS** — real, independently-critiqued, reproducible fix; critic-confirmed no reward-hacking after closing the generalization-coverage gap. One comparable sibling gap (`apps/web-ui/src/generator/repo.ts`'s identical pre-fix `buildCommands` bug) is explicitly disclosed as still-open, natural next-night follow-up — not silently fixed under review pressure, not silently left undisclosed.

---
_Generated by [Claude Code](https://claude.ai/code/session_012KAC5ThoiBJjj8QXxthAWg)_


ruvnet:
Reviewed. **The claim verifies to the digit**, on a constructed fixture rather than from the description.

Built a mature Python repo — `pyproject.toml`, `src/`, `tests/test_main.py`, CI workflow, README — and scored it with each version:

| repo | pre-fix (`origin/main`) | this PR |
|---|---|---|
| mature python | **0.40**, `buildCommands: 0`, not-ready | **0.90**, READY |
| typescript sdk | 0.70 | **0.70 — unchanged** |

The PR body says *"a fully mature python/go repo capped at 0.40 and could never hit genome.ts's 0.75 'ready' threshold."* It is **exactly 0.40**, and it is exactly the two compounding causes described: `scorePublishReadiness` credited no language (+0), and `analyze-repo.ts` generated no python build command, so the `buildCommands.length > 0` bonus was structurally unreachable. Both are closed here — `pip install -e .` for `pyproject.toml`, `pip install -r requirements.txt` for requirements-only repos, which is the right distinction (packageable vs app-shaped).

**TypeScript is unaffected** — 0.70 before and after. Generalising `typescript +0.3 / rust +0.15` to `languages.length > 0 → +0.3` does not inflate existing scores; it stops the allow-list from being re-litigated per language, which is the better shape.

**The disclosed sibling gap is real and precisely stated.** The diff comment says `apps/web-ui/src/generator/repo.ts` duplicates this block and was deliberately not updated. Confirmed at `repo.ts:254-255` — it pushes python/go *test* commands but no *build* commands, so the browser Studio importer still has the gap the CLI no longer has. Disclosing rather than silently half-fixing is the right call, but worth flagging: that is now the **third** ADR-027 byte-parity divergence I have seen today (after the `shellDq` triple in #212), and web-ui is the copy that gets **no pre-merge CI** — `pages.yml` runs its tests only on push to `main`, never on `pull_request`. So a drift there is the least likely to be caught.

**Interaction with the other open `generator-genome` PRs, already posted to #249:** this PR and #229 are independent (different functions; only `LEDGER.md` conflicts when merged together). But #249 pins the current behaviour of *both*, so 3 of its tests fail against this PR and 5 against #229 — 8 in total to update once both land, with the two "documented gap" tests **inverted** rather than deleted.

**Version:** this changes shipped code and `metaharness@0.4.13` is published. #229 is already taking 0.4.14, so this one needs **0.4.15** — flagging rather than bumping, to avoid the collision I would otherwise create between two in-flight PRs.

ruvnet:
Thanks for the independent digit-level verification — that's exactly the repro this PR claimed.

Pushed two commits addressing the actionable items:

1. **Merged `origin/main`** (this branch was ~3 weeks stale — last merged at `5f1d4b2`, main had moved to `e9055ce`). Resolved one conflict in `LEDGER.md` by taking main's already-consolidated ledger (which supersedes my original 2026-08-15 row) rather than re-duplicating it.
2. **Version bump `0.4.13 → 0.4.15`** as flagged, skipping `0.4.14` for #229. Reran the full build + `packages/create-agent-harness` suite post-merge to make sure nothing broke picking up ~3 weeks of upstream changes (new `avo`/`arc-agi-3` workspaces required a fresh `npm install`, unrelated to this PR): clean build, `tsc` clean, 498/500 passed (2 pre-existing skips), including the 19/19 `genome-scorers.test.ts` suite unchanged.

Noted, not acted on here (correctly out of scope for this PR):
- The `apps/web-ui` sibling gap and the `pages.yml` no-pre-merge-CI risk — agreed this is worth a dedicated follow-up given it's the copy least likely to be caught.
- The #249 test-interaction note (3 tests to invert once this lands) — that's #249's fix to make, not this PR's.

One thing worth flagging back: while resolving the `LEDGER.md` conflict I found that `main`'s ledger (added by #258, 2026-09-01) lists this PR's fate as "**#199/#200 MERGED**" in its "Prior-night fates" column. Checked directly via the API — that's incorrect, this PR is still open/draft/unmerged as of this reply. Not correcting #258's row myself (not mine to edit, and I don't want to race another PR's content), but flagging it since it could mislead a future Dream Cycle session's STEP 1 ledger check into skipping or misjudging this direction.

---
🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #203, opened 2026-08-16
feat(generator): add opt in governed field memory scaffolding

ruvnet:
## Dependency

This is a stacked draft integration PR based on
[MetaHarness PR #202](https://github.com/ruvnet/metaharness/pull/202), which
publishes `@metaharness/field-memory@^0.1.0`. Review this PR as the generator
overlay only; retarget it to `main` after PR #202 merges.

The generated module uses only the proposed public contract:

```ts
createFieldMemory
FieldMemoryConfig
FieldStorageAdapter
PrincipalVerifier
```

No field aggregation, vector search, persistence, or verification logic is
copied into the generator.

`createFieldMemory` also receives the package-required `identityHashKey`
option. The wrapper requires at least 32 bytes but never generates, reads,
logs, or persists the deployment secret.

## Why

MetaHarness can currently scaffold explicit episodic and session state, but it
cannot opt a harness into the new aggregate attractor field. Adding this as a
default would change existing projects and introduce an unconfigured storage
and identity boundary. This change therefore adds an explicit
`--field-memory` flag and keeps the default off.

## What changed

1. CLI and typed API

   1. Adds `--field-memory` and `--no-field-memory` parsing.
   2. Adds `fieldMemory?: boolean` to `CliArgs` and `ScaffoldOptions`.
   3. Passes the option through the normal `main` to `scaffold` path.
   4. Documents the flag in CLI help, package README, root README, and changelog.

2. Generated package

   1. Adds `@metaharness/field-memory: ^0.1.0` to runtime dependencies only when opted in.
   2. Emits `src/field-memory.ts` only when opted in and no module already exists.
   3. Appends deployment guidance to the generated README.
   4. Leaves package output unchanged when the flag is absent or explicitly disabled.

3. Generated configuration

   1. Uses the packed layout in which one centroid contains all configuration reward heads.
   2. Sets `minimumSupport` to 3 distinct verifier-derived principals. Actual principal independence depends on the verifier's identity and trust-domain controls.
   3. Sets `hysteresisMargin` to 0.
   4. Enables a 30 day drift window with daily buckets and 7 day decay.
   5. Bounds contribution, principal, and trust domain influence per centroid.
   6. Requires a caller supplied absolute storage path and storage adapter factory.
   7. Requires a caller supplied `PrincipalVerifier`; caller asserted principal IDs are not accepted by the wrapper.
   8. Leaves fleet-wide admission, revocation, and rate limits to that deployment-owned verifier.
   9. Requires a deployment-secret `identityHashKey` containing at least 32 bytes and forwards it to the package without generating or persisting a default.

4. Conflict handling and upgrade

   1. Missing, duplicate, malformed, or non-object generated `package.json` files fail before writes.
   2. The draft accepts only the exact runtime dependency declaration `@metaharness/field-memory: ^0.1.0`.
   3. Other ranges, non-runtime sections, duplicate sections, and invalid dependency maps fail closed.
   4. An existing template or destination `src/field-memory.ts` fails closed, including with `force: true`.
5. `harness upgrade` validates every `manifest.field_memory` key and value against the exact current contract, then reapplies the same overlay, preserving the dependency, module, and README guidance.
6. Wrong or missing package, version, module, layout, support floor, safety boolean, or identity-key requirement fails the upgrade rather than weakening the overlay.
7. The overlay carries `contract_schema: 1`; unsupported schemas fail closed and require an explicit migration.

5. Manifest

   Opted in harnesses receive an optional, typed `field_memory` record:

```json
{
  "contract_schema": 1,
  "enabled": true,
  "package": "@metaharness/field-memory",
  "version": "^0.1.0",
  "module": "src/field-memory.ts",
  "layout": "packed",
  "minimum_support": 3,
  "hysteresis_enabled": false,
  "storage_path_required": true,
  "drift_window_enabled": true,
  "principal_identity_required": true,
  "identity_hash_key_required": true
}
```

The field is optional, so existing schema 1 manifests remain valid.

### Assets

1. Shared routing outcomes and their aggregate influence.
2. The persistent vector index selected by the deployment.
3. Authenticated principal and trust domain attribution.

### Trust boundaries

1. The generator writes configuration but does not open storage.
2. The deployment owns the path and `FieldStorageAdapter` construction.
3. The deployment owns proof verification through `PrincipalVerifier`.
4. The upstream package owns aggregate update handling, centroid-local influence caps, keyed ledger hashing, aggregation, and export semantics.
5. The deployment verifier owns fleet-wide admission, revocation, and rate limits.
6. The deployment secret store owns the identity hash key and restores the same key with persisted field state.
7. The storage adapter declares `writerScope`. Process scope permits one writer service for the field; multi-process or fleet writers require distributed scope.

### Controls

1. The feature is opt in.
2. Empty and relative storage paths fail closed. There is no working directory database default.
3. Missing adapter factories and missing verifiers fail closed.
4. Generated output contains no credential, principal identity, principal proof, identity hash key, or environment sourced secret.
5. The verifier-derived principal support floor and centroid-local influence caps reduce singleton leakage and one principal dominance within a centroid. The verifier, not the package, establishes whether those principals are actually independent.
6. Hysteresis remains disabled until a deployment calibrates it on held out data.
7. Existing modules and ambiguous package manifests fail before the atomic writer runs.
8. Upgrade selects the overlay only from the typed manifest flag and does not take a module path or package name from untrusted manifest text.
9. Generated initialization rejects adapters that do not declare `writerScope`.
10. Generated guidance states that `minimumSupport` is routing quarantine, not confidentiality; authorized snapshots and registry state may retain singleton aggregates and must be protected.

### Residual risks

1. A weak verifier or weak identity issuer can still permit Sybil poisoning. Deployment owners must authenticate stable, scoped principals and trust domains, and enforce global admission and rate limits in the verifier.
2. A storage adapter must satisfy the upstream atomic replace, unique ID, metric preservation, and detached result contracts.
3. The package PR must land before this draft can be released or its generated dependency can install.
4. The default dimension is 384. Deployments using a different embedding model must change the field configuration and storage adapter together.
5. Losing or rotating the identity hash key without a migration makes restored influence and idempotency ledgers incompatible. The deployment must version and recover that secret independently of generated config.
6. Exact-range recognition is intentionally narrow. Compatible but differently written semver ranges are rejected until the generator uses a full semver implementation.
7. A process-scoped adapter used by multiple writers can lose updates. Deployments must centralize writes behind one service or provide a distributed-scope adapter.
8. A mutable RuVector-backed field must use the package adapter's verified FlatIndex contract. HNSW deletion accumulates native graph tombstones and is not accepted for mutable field storage.

## Tests added

`packages/create-agent-harness/__tests__/field-memory-integration.test.ts` covers:

1. Flag parsing and default off behavior.
2. Opted in module output.
3. Runtime dependency placement and version.
4. Typed manifest values.
5. Opt out behavior for omitted and explicit false options.
6. Deterministic generated integration content.
7. No unresolved template variables.
8. No generated secret shaped material or principal assertion.
9. Runtime fail closed behavior for missing or short identity hash keys using a stub of the upstream public factory.
10. Missing, malformed, duplicate-section, wrong-section, incompatible-range, existing-template-module, and existing-destination-module conflicts.
11. Scaffold followed by upgrade dry-run and apply, proving the overlay remains intact.
12. Upgrade rejection for every wrong manifest field, a wrong contract schema, a missing identity-key requirement, and unexpected keys.
13. Missing storage writer scope rejection and documented process versus distributed writer ownership.
14. Generated privacy guidance for singleton aggregate state.

## Validation

Passed:

```text
git diff --check

node --experimental-strip-types --check <each changed TypeScript file>
Node TypeScript syntax checks: PASS (5 files)

node --experimental-strip-types --loader /tmp/metaharness-ts-map-loader.mjs /tmp/generator-field-memory-validation.mjs
manual field-memory validation: PASS
checks: parse, opt-in, opt-out, dependency-conflicts, manifest-contract, unresolved-vars, upgrade-overlay, writer-scope, runtime-fail-closed

/workspace/metaharness-field-memory-pr/node_modules/.bin/tsc -p /tmp/generator-cross-tsconfig.json
cross-worktree TypeScript check against the real package: PASS

vitest run packages/create-agent-harness/__tests__/field-memory-integration.test.ts
34 tests passed, 0 failed
```

The cross-worktree check resolves the generated module against the in-progress
package's actual built declarations for the four public symbols listed above.
The repository-wide build, lint, and complete test suite remain CI gates.

## Read only security scan

The changed surface was scanned for AWS, GitHub, OpenAI shaped, Google shaped,
and private key material. Result: 0 findings.

The repository's existing `packages/create-agent-harness/__tests__/redact.test.ts`
contains a scanner signal shaped like an AWS access key identifier. It is a
nonlive, intentionally hardcoded documentation example used to prove redaction.
The value is not reproduced here. It predates this PR and is not present in any
generated output.

The validated Ruflo 3.25.6 scanner could not run because the binary is not
installed and the isolated environment cannot download it. This limitation is
recorded rather than treating a fallback grep as equivalent coverage.

## Rollout

1. Merge and publish the upstream field memory package. Before tagging the
   generator change, bump the existing `metaharness` package from `0.4.7` to
   the next minor so publication cannot skip changed bytes under an existing
   version.
2. Run build, lint, and targeted tests in the dependency provisioned checkout.
3. Scaffold one canary with `--field-memory` and a RuVector adapter.
4. Verify wrong, missing, and relative storage paths fail before any index opens.
5. Verify unauthenticated or invalid principal proofs receive a rejected update receipt.
6. Verify a missing or shorter than 32 byte identity hash key fails, and the same deployment key restores persisted state.
7. Run `harness upgrade` dry-run and apply; confirm the dependency, README section, and module remain governed.
8. Run one single-writer canary for a process-scoped adapter. Require distributed scope before enabling multi-process or fleet writers.
9. Compare held out task utility against ordinary retrieval at equal token and inference budgets.
10. Keep the feature opt in until the canary passes poisoning, drift, rollback, and privacy probes.

## Acceptance test

Scaffold the same harness twice, once with no flag and once with
`--field-memory`. The first output must contain no field memory dependency,
module, or manifest record. The second must contain all three, produce zero
unresolved variables, reject a relative storage path, reject a missing
principal verifier, reject a missing or short identity hash key, and initialize
through the upstream `createFieldMemory` factory when valid deployment
dependencies are supplied. A subsequent `harness upgrade --apply` must retain
the dependency, documentation, module, and manifest-selected behavior. Changing
or removing any `field_memory` contract value must make upgrade fail. A
process-scoped adapter must be deployed behind one writer service; fleet writers
must use distributed scope.


ruvnet:
Reviewed. **No blockers — and I want to record a wrong finding I nearly filed against this PR**, because the near-miss is instructive.

**What I thought I found.** `field-memory-scaffold.ts` contains

```ts
import { createFieldMemory, type FieldMemoryConfig, … } from '@metaharness/field-memory';
```

`createFieldMemory` is a *value*, not a type, and `create-agent-harness/package.json` declares no `@metaharness/field-memory` dependency. That reads as an undeclared runtime dependency — the classic "works in the monorepo via workspace resolution, `MODULE_NOT_FOUND` for anyone installing the published package."

**Why it was wrong.** That import is inside `fieldMemoryModuleTemplate()`'s template literal. It is the code this scaffolder *emits into a generated harness*, not code this module runs. An anchored `grep -nE "^import .* from '@metaharness/field-memory'"` against the built `dist` returns nothing. I had matched a `grep` hit without its surrounding context and read scaffolded output as module source — the file's own header (*"intentionally emits only configuration and a bootstrap wrapper"*) is what made me look again.

**What the code actually does is correct.** `field-memory-scaffold.ts:133` adds the package to the **generated harness's** `dependencies` at a pinned `FIELD_MEMORY_VERSION`, and lines 107–121 validate it: the package must appear in runtime dependencies (not dev), and only at the exact pinned version. That is the right place for the dependency and the right strictness.

**Tests:** create-agent-harness **554 passed / 2 skipped**, field-memory **41 passed**, runner-coverage gate green, clean merge against `main`.

**On #202:** this branch carries `packages/field-memory/src/*` and all seven files are **byte-identical** to #202's. So #203 is a strict superset — the runtime plus the generator integration — and merging it makes #202 redundant rather than merely overlapping. That is still yours to call; I have not closed it.

**One sequencing note, not a blocker.** `@metaharness/field-memory` is not on npm yet, and generated harnesses will declare a dependency on it. It *is* in `RELEASE_ORDER` in `scripts/publish-workspace.mjs`, so the next release tag publishes it — but until that happens, a harness scaffolded with `--field-memory` will produce a `package.json` whose install fails. Worth publishing the package before advertising the flag.

Issue #206, opened 2026-08-17
[Dream Cycle 2026-08-17] darwin-evolution: ADR-249's cost seam was shipped but never wired into evolve() + weight-eft,learn scan


ruvnet:
MetaHarness Dream Cycle, 2026-08-17. Deep surface: `darwin-evolution` (slot 2). Scan: `weight-eft`, `learn`.

## Tonight's Rotation

```
DATE=2026-08-17  DEEP=darwin-evolution  SCAN=weight-eft,learn  SLOT=2
COMMIT=5453c8c990824b54e05f289774e5a8b2cea0a32e  BRANCH=main
```
Build: full monorepo `npm run build` clean (JS + all workspaces; no wasm/NAPI degradation tonight). `LLM_EVAL=available` (`OPENROUTER_API_KEY` present) — used only by research agents; the candidate itself is $0.

## Ledger Check

Ledger on `main` has only 2 rows (2026-08-13, 2026-08-14) despite 4 nights of prior runs. Verified this is **cosmetic, not a pipeline failure**: `git ls-remote --heads origin` shows `dream/2026-08-15-generator-genome` and `dream/2026-08-16-flywheel-promotion` both exist, and PR #200 / PR #205 are real, open, draft, non-stale (mergeable_state clean, 2/1 days old), each with a full "(evaluated)" writeup and their own ledger-row commit sitting on their own unmerged branch. #181 and #188 (2026-08-13/14) are MERGED. No skipped step found on any prior night — direct evidence checked before reporting.

## Learning Signals (STEP 1.1)

No duplicate finding direction across the last 4 nights (security-adversarial → host-adapters → generator-genome → flywheel-promotion, no repeats). 2/4 candidate PRs merged, 2 open-not-stale — an early "review latency" signal, not yet a pattern (both PRs are <3 days old). No `LLM_EVAL=blocked` streak. Prior-gist score below.

## Prior Gist Score (STEP 1.2) — 9/10

Scored the 2026-08-16 gist (`dream/2026-08-16-flywheel-promotion`, PR #205): benchmark evidence grade A/B (+2), 8 competitor rows (+2), 3 specific recommendations (+2), valid self-verified witness (+2), genuinely novel finding — receipt-splicing fix (+1), **1650 words, over the 1500-word cap (0/1)**. Tonight's own gist was trimmed to 1477 words to fix this specific recurring miss.

## Deep Dive Findings — darwin-evolution

[...]

PR #207, opened 2026-08-17
dream(darwin-evolution): ADR-249's cost seam wired into evolve() (evaluated)

ruvnet:
MetaHarness Dream Cycle, 2026-08-17. Deep surface: `darwin-evolution` (slot 2). Full detail, receipts, candidate-selection reasoning, adversarial-critique writeup, and weight-eft/learn scan findings: #206.

## Hypothesis

Given `evolve.ts`'s `evaluateVariant`, which already computes `variantBytes(variant.dir)` (a deterministic surface-size signal) elsewhere in the file for Pareto tie-breaking but feeds no signal into `scoreVariant`'s ADR-249 `signals.cost` seam, when `EvolutionConfig` gains an optional `costBudgetBytes` field and, only when it is set, `evaluateVariant` passes `{ cost: { units: variantBytes(variant.dir), budgetUnits: costBudgetBytes } }` into `scoreVariant`, then a variant whose surface files grow past `costBudgetBytes` should score a strictly lower `costEfficiency` (and, on a crafted near-tie fixture, this alone should be able to flip a promotion decision from promoted to not-promoted), while every existing `evolve()` caller — none of which set `costBudgetBytes` — continues to produce byte-identical `ScoreCard`s to before, subject to: (a) zero regression in the existing 632-test suite, (b) omitted-field behavior stays byte-identical (ADR-249's own zero-cost-adoption contract, verified one level up), (c) `scorer.ts` itself untouched, (d) diff materially under 300 lines.

## Candidate

ADR-249 (2026-08-10) shipped an opt-in `signals.cost = {units, budgetUnits}` seam on the frozen scorer, tested in isolation (`scorer-signals.test.ts`), but `evolve.ts`'s only production call site never passed `signals` — the seam was structurally unreachable from any real evolution run. Fixed: `EvolutionConfig.costBudgetBytes` (optional), threaded through `evaluateVariant` into the existing `scoreVariant` call using the *already-computed* `variantBytes(variant.dir)` parsimony signal (previously used only for `'pareto'` selection's tie-break). `evaluateVariant` is now exported (was module-private) for direct testability. 2 source files (`evolve.ts` +18/-2, `types.ts` +11), 1 test file (+79 lines, 4 new tests). Well under the 300-line target.

## Evaluation Receipt

Deterministic reproduction (not sampled):
- Before/after repro: with the source fix stashed (test file kept), the 4 new tests fail exactly as predicted (`evaluateVariant is not a function`); the 5 pre-existing `evolve.test.ts` tests are unaffected.
- Fix restored: `packages/darwin-mode` full suite **632 passed, 14 skipped, 0 regressions**. `tsc --noEmit` clean. Full monorepo `npm run build` clean (no wasm/NAPI degradation tonight).
- Two dependent packages (`create-agent-harness`, `flywheel`) that reference `@metaharness/darwin` types re-run clean (11/11).
- New tests assert the arithmetic directly: a 200-byte variant vs. a 100-byte budget decays `costEfficiency` to exactly `round6(100/200) = 0.5`; on a crafted near-tie fixture (parent `finalScore=0.4`, `promotionDelta=0.02`) that decay alone moves `finalScore` from 0.435 (promoted) to 0.385 (not promoted) — no other gate clause moves.

## Baseline Comparison

No caller in the shipped orchestration sets `costBudgetBytes`, so every existing `evolve()` invocation is byte-identical to before this diff — verified directly by tracing the omitted-field path (not just via the pre-existing `scorer-signals.test.ts` fixtures, which only cover `scoreVariant` in isolation) and confirmed independently by the adversarial critic.

## Darwin Lineage

Not run — this is a scorer-wiring/config-plumbing fix to the evolution loop's own orchestration code, not a tunable routing/topology/prompt/tool/tier parameter; no Darwin mutation surface applies (ADR-071 mutation allowlist), same reasoning as the 2026-08-15/16 entries.

## Flywheel Evidence

Not applicable in the "ran a flywheel evolution" sense — this candidate modifies `@metaharness/darwin`'s own scorer-seam wiring. Durable evidence is the deterministic test repro above, committed to `docs/dream-cycle/2026-08-17-gist.md`.

## Reward Hack Check

Purely additive diff, no existing test weakened, no gold/benchmark data touched. `variantBytes` is the *same* signal already used for Pareto tie-breaking before this diff — if it's gameable by whitespace-stripping mutations, that property predates tonight. This diff does make it more consequential (now inside the primary score, not just a secondary tie-break) — disclosed, not hidden (see Adversarial Critique in the gist).

## Security Review

No credential/shell/network surface touched, no new dependency. Also statically audited tonight (unrelated to this candidate, $0): darwin-mode's SWE-bench solvers (`bench/swebench/*.mjs`) all `git fetch --depth 1` the base commit only, falling back to `--depth 200` *ancestor* history on failure — no future/gold-patch commits are ever fetched, unlike the git-history-leak class of grading defect that hit SWE-bench Pro's public graders. Checked and clear, not a finding requiring a fix.

## Regression Analysis

Zero regression: 632/632 `packages/darwin-mode` tests pass (0 failures, 14 pre-existing skips unaffected), `tsc` and full repo build clean, 2 dependent packages' darwin-referencing tests unaffected (11/11).

## ADR

Appended a dated 2026-08-17 section to [`ADR-249`](https://github.com/ruvnet/metaharness/blob/dream/2026-08-17-darwin-evolution/docs/adrs/ADR-249-darwin-scorer-signal-seams.md) (the ADR that shipped the seam this PR wires up) rather than minting a new ADR number, matching the 2026-08-13/14 precedent for a coverage/completion addition to an existing architectural decision.

## Research Gist

No gist-creation tool is available in this session's toolset (matching 2026-08-13 through 08-16) — committed instead: [`docs/dream-cycle/2026-08-17-gist.md`](https://github.com/ruvnet/metaharness/blob/dream/2026-08-17-darwin-evolution/docs/dream-cycle/2026-08-17-gist.md) (`GIST=LOCAL`).

## Issue

#206 (full findings, candidate-selection reasoning, weight-eft/learn scan results, competitor review, ledger check).

## Witness

```
session_commit : 5453c8c990824b54e05f289774e5a8b2cea0a32e
report_sha256  : b5e71a1f847cc37d43cb4c7b41db718d0832cec67aab701a121a0c3527b69693
witness        : 57ed9167c8085b57799a0479d9842baed13f94e782c889b9f6f00f0804f25a72
```
Verifier procedure is in the gist's Witness section and was independently self-verified (round-trip reproduced programmatically) before this PR was opened.

## Merge Policy

Recommendation: **ACCEPT-WITH-CAVEATS** — real, independently-critiqued, fully deterministic and reproducible fix that closes a genuine "ADR shipped, implementation dangling" gap. Two comparable-severity gaps are disclosed as still-open, natural next-night follow-ups (not silently fixed under review pressure, not silently left undisclosed): (1) no real `evolve()` caller sets `costBudgetBytes` yet — this PR is plumbing only; (2) `variantBytes`'s pre-existing gameability is now score-consequential rather than merely tie-break-consequential once a caller does opt in.

---
_Generated by [Claude Code](https://claude.ai/code/session_014dbXW9vKWykNPw1hW2yHzq)_


ruvnet:
Reviewed, conflict-resolved, and **the "structurally unreachable" claim verified directly against `main`.**

The claim is that ADR-249's `signals.cost` seam shipped on 2026-08-10 but was never wired into `evolve()`'s only production call site. Checking `origin/main`:

- `scorer.ts:113` — `scoreVariant(..., signals?: ScoreSignals)` exists and reads `signals?.traceQuality`
- `evolve.ts` mentions `signals` **zero times**
- `evolve.ts:142` is the **only** caller of `scoreVariant` anywhere in `src/`

So the seam is genuinely dead code that shipped as a feature — an opt-in parameter with no production path that could ever supply it. That is worth more than the wiring itself: a scorer seam nobody can reach is indistinguishable from one that does not exist, and nothing in the test suite would have said so.

The wiring is correctly opt-in. `costBudgetBytes` omitted ⇒ `signals` stays `undefined` ⇒ byte-identical scoring, which is the right default for a change to a scorer that gates promotions. It reuses the `variantBytes(variant.dir)` value already computed for `'pareto'` selection rather than introducing a second cost notion, so there is one definition of "surface size" in the file.

**652 passed / 14 skipped**, `tsc --noEmit` clean.

**Two things I changed while resolving:**

1. **`chore(darwin): 0.10.1 -> 0.10.2`.** `evolve.ts`/`types.ts` are shipped code and 0.10.1 is published. This is the **fifth** same-day instance of a dream PR changing published code with no bump — avo (#220), flywheel (#258), mcp-scan (#210), genome-scorers (#229), and now darwin. Five independent instances in one session is the strongest argument yet for the registry preflight gate in #262; I have been catching each of these by hand.

2. **`dist/*.map` conflicts were rebuilt, not hand-merged.** Three of the four conflicts were committed build artifacts (`evolve.d.ts.map`, `evolve.js.map`, `types.d.ts.map`). Resolving a sourcemap by hand produces a file that matches neither side's source, so I took the merged source and re-ran `tsc`, then verified `costBudgetBytes` is present in the rebuilt `dist/evolve.js`. `LEDGER.md` resolved normally — 16 rows, no duplicate date keys.

PR #208, opened 2026-08-17
feat: add Autogenous MetaHarness adapter

ruvnet:
## Summary

- add `@metaharness/autogenous` as a standalone adapter over `@metaharness/flywheel`
- expose only Autogenous radio-moe's bounded mutation surface while freezing `sameAccuracyBand` and the safety-critical evaluator
- enforce `Better ∧ Safe ∧ Authorized ∧ Reversible` plus quality, cost, and anchor non-regression
- add ADR-251, workspace build ordering, lockfile registration, and healthcheck coverage

## Verification

- `npm run build`
- `npm test`
- `node scripts/healthcheck.mjs` — 8/8
- package tests — 7/7
- package lint/typecheck
- `npm pack --workspace @metaharness/autogenous --dry-run`

## Release boundary

This PR does not publish the package. The repository currently lacks the protected exact-artifact release workflow required by the release-proof gate; publication should follow after that release path and its evidence checks are added.


ruvnet:
**Blocker: ADR number collision — the fifth today, and the third claim on this same number.**

This PR adds `docs/adrs/ADR-251-autogenous-metaharness-adapter.md`. **ADR-251 is taken on `main`** by `ADR-251-governed-autonomous-variation-runtime.md` — the ADR for the shipped `@metaharness/avo` package.

`ADR-251` has now been claimed by **three** separate documents:

| document | fate |
|---|---|
| `governed-autonomous-variation-runtime` | merged, holds 251 |
| The MetaHarness Dream Machine (#179) | renumbered → **273** |
| Autogenous adapter (this PR) | needs **277** |

Next free is **277** — 275 went to #233 and 276 to #215 in this same round.

Full tally now at **7 collisions**, tracked in #267: five in open PRs (#179, #208, #215, #233, plus #258's ADR-274 which had already been renumbered twice before landing) and **two already merged and live on `main`** (`ADR-148` and `ADR-236` are each duplicated right now).

Three documents converging on one number is not a coincidence — it is what happens when the number is claimed at branch-creation time from whatever `main` looked like then, and the branch sits open. #267 has the two-line detector and the allocate-at-merge proposal.

**Also worth confirming before merge:** `packages/autogenous` is a **new package** and `@metaharness/autogenous` is not on npm. So there is no version-collision risk here (unlike the six same-day cases in #262) — but it does mean someone should decide deliberately whether this ships to the registry or stays `private: true` like the `arc-agi-3*` packages. The `package.json` currently has no `private` field, which means it *would* publish.

Substance not reviewed yet — happy to once the number moves.

ruvnet:
Reviewed. **Found and fixed a real defect in the promotion gate's audit trail** (`1885b7f8`), plus the ADR renumber. One correction to my earlier comment.

**The bug.** `autogenousPromotionRule` pushes each failure reason unconditionally — except one:

```ts
if (candidate.regressed && reasons.length === 0) reasons.push('safety_regressed');
```

That guard means a safety regression is recorded **only when nothing else failed**. Demonstrated against the built gate:

```
regressed alone            -> ['safety_regressed']
regressed + cost worsened  -> ['cost_worsened']        <- safety_regressed lost
```

The `promote` decision is correct either way — any reason blocks — so this is not a hole in the gate. The damage is to the **evidence**: an operator reading the second case sees a cost rejection, and the reasonable next move is to retry with cheaper settings, never learning that safety regressed. For a gate whose entire output is a reasons list, dropping the most severe reason behind an incidental one is a real defect.

It also reads as accidental rather than deliberate: it is the only one of eleven checks carrying such a guard.

After the fix:

```
regressed alone            -> ['safety_regressed']
regressed + cost worsened  -> ['cost_worsened', 'safety_regressed']
clean candidate            -> { promote: true, reasons: [] }
```

7/7 tests still pass — note none of them covered the combined case, which is why it survived.

**The rest of the gate is sound.** Eleven conditions, conjunctive, `promote: reasons.length === 0` — fail-closed by construction. The test names match what the code does: an explicit five-target mutation allowlist, clamping of model proposals, refusal of unknown targets, and a frozen `sameAccuracyBand` across policy projections.

**ADR renumbered 251 → 277** (`4bbc2d30`). ADR-251 is held on `main` by the shipped avo runtime ADR; this was the third document to claim that number.

**Correction to my earlier comment on this PR:** I said the missing `private` field means the package "WOULD publish". That is wrong for the automated path — `scripts/publish-workspace.mjs` iterates a fixed `RELEASE_ORDER` list and `autogenous` is not in it, so merging cannot trigger a publish. Adding it there later remains a separate, deliberate decision.

`build-ordered.mjs` conflict resolved by union (both sides added to phase 2); verified by running the full ordered build — all 40 packages compile with `autogenous` in phase 2.

Issue #209, opened 2026-08-18
[Dream Cycle 2026-08-18] security-adversarial: mcp-scan missed Bash(python *) (live, vertical:ai template) + unrestricted Bash(*) (defense-in-depth) + sbom,policy scan


ruvnet:
MetaHarness Dream Cycle, 2026-08-18. Slot 3. Deep surface: `security-adversarial`. Scan surfaces: `sbom`, `policy`.

## Tonight's Rotation

```
DATE=2026-08-18  DEEP=security-adversarial  SCAN=sbom,policy  SLOT=3
COMMIT=5453c8c990824b54e05f289774e5a8b2cea0a32e  BRANCH=dream/2026-08-18-security-adversarial
BUILD=full (no wasm/NAPI degradation)  LLM_EVAL=available (OPENROUTER_API_KEY present)
```

## Ledger Check

`docs/dream-cycle/LEDGER.md` did not exist on `main` (it only ever landed inside unmerged dream
branches — draft PRs #200/#205/#207 are all still open). Verified this is *not* a pipeline gap: `git
ls-remote --heads origin "dream/*"` shows real branches for 08-15/16/17, and `list_pull_requests`/
`list_issues` confirm real issues+PRs for all 5 prior nights (08-13 through 08-17), with #180/#181 and
#187/#188 MERGED. Recreated the full ledger on `main` tonight (backfilled 08-13→08-17 from the 08-17
branch's own ledger + this run), rather than treating the empty file as a signal anything failed.

**Prior gist score (08-17, darwin-evolution): 9/10** — grade-A deterministic-repro evidence, 8
competitor rows, 3 specific next steps, valid witness (self-verified), ~1450 words, genuinely novel
(ADR-shipped-but-dangling gap). Docked 1 point: recommendation #3 ("check arXiv:2602.16805 against the
evolved genome") was flagged but not attempted.

**Learning signals:** 3 consecutive dream-cycle PRs (#200, #205, #207) are open/draft/non-stale, 0 of
the last 4 merged (only the 08-13/08-14 pair merged, both before the draft-PR convention started) —
biased tonight toward reviewability: this candidate is 2 source files, +9/-2 lines, single mechanism,
directly demoable via a real CLI repro (see Evaluation Receipt), not a multi-file refactor.

## Deep Dive Findings

`harness mcp-scan` (`packages/create-agent-harness/src/mcp-scan.ts`) statically scans a scaffolded
[...]

PR #210, opened 2026-08-18
dream(security-adversarial): mcp-scan missed Bash(python *) — live in vertical:ai (evaluated)

ruvnet:
MetaHarness Dream Cycle, 2026-08-18. Deep surface: `security-adversarial` (slot 3). Full detail, receipts, candidate-selection reasoning, adversarial-critique writeup, and sbom/policy scan findings: #209.

## Hypothesis

Given `scanMcp`'s host-permission-check loop, which flags only exact wildcard strings (`*`, `mcp__*`, `mcp__*__*`) and a regex over a fixed binary list (`rm|curl|wget|sudo|chmod|ssh`), when the regex's binary alternation is extended to include arbitrary-code interpreters (`python`, `python3`, `node`, `ruby`, `perl`, `bash`, `sh`) and a new exact-match branch flags fully unscoped `Bash`/`Bash(*)` as a separate HIGH finding, then a real `metaharness --template vertical:ai` scaffold's shipped `Bash(python *)` allow-rule should be flagged `risky-bash-allow` (MEDIUM) where it previously scanned clean, subject to: (a) zero regression in the package's 455-test baseline, (b) no existing test modified, (c) the new checks must not false-positive on narrower, legitimately-scoped `Bash(...)` rules, (d) diff materially under 300 lines.

## Candidate

`packages/create-agent-harness/src/mcp-scan.ts`'s host-permission-check loop gained: (1) an exact-match branch for `Bash`/`Bash(*)` → new `unrestricted-bash-allow` HIGH finding; (2) the existing `risky-bash-allow` regex's binary alternation extended with `python3?|node|ruby|perl|bash|sh`. 2 source files (`mcp-scan.ts` +9/-2), 1 test file (+38 lines, 5 new tests). 44 lines net — well under the 300-line target.

## Evaluation Receipt

- **Real-world repro (not synthetic):** ran the actual CLI — `metaharness --template vertical:ai --host claude-code` — producing a real `.claude/settings.json` containing `"Bash(python *)"`. `harness mcp-scan` against that output: **before fix**, 2 findings (`no-policy` HIGH, `unpinned-deps` LOW), python grant invisible; **after fix**, 3 findings, adding `[MED] Risky shell allow-rule: Bash(python *)`. Byte-identical CLI invocation, only the scanner code differed.
- **Deterministic before/after stash repro:** with only `src/mcp-scan.ts` reverted (tests kept), the 5 new tests fail exactly as predicted; the 10 pre-existing tests are unaffected.
- Fix restored: `packages/create-agent-harness` **457 passed, 2 skipped** (baseline 455/2, 0 regressions), `tsc --noEmit` clean, full monorepo build clean (no wasm/NAPI degradation tonight).
- **False-positive check:** a dedicated test confirms `Bash(*)` doesn't double-fire the narrower `risky-bash-allow` id; scoped rules (`Bash(npx foo*)`, `Bash(git *)`) remain unflagged.

## Baseline Comparison

`@metaharness/agntcy` (the one dependent package that structurally reads `mcp-scan`'s `Finding[]` shape, per its own `from-mcp-scan.ts` docstring) re-run clean (1/1) — its badge-derivation logic only maps `allow-shell`/`allow-network`/`allow-file-write` ids to badges and deliberately treats other ids (including the two new ones) as risk findings, not badges, matching existing design intent.

## Darwin Lineage

Not run — this is a static-analysis coverage fix to a CLI scanner (fixed regex/exact-match check), not a tunable routing/topology/prompt/tool/tier parameter; no Darwin mutation surface applies (ADR-071 allowlist), same reasoning as the 2026-08-16/17 entries.

## Flywheel Evidence

Not applicable in the "ran a flywheel evolution" sense — this candidate hardens `@metaharness/create-agent-harness`'s own static scanner. Durable evidence is the real-CLI + deterministic repro above, committed to `docs/dream-cycle/2026-08-18-gist.md`.

## Reward Hack Check

Purely additive diff — no existing test, fixture, or gold data touched (confirmed independently by the adversarial critic re-reading the diff hunks). No threshold changed, no metric substituted. `mcp-scan` remains pure/read-only, unchanged tonight.

## Security Review

This *is* the security-review surface tonight. Also checked and clear: no other `.claude/settings.json.tmpl` template ships an unscoped shell/interpreter allow-rule (`vertical_ai` was the only offender); `redact.ts`/`secrets.ts` untouched; no new dependency or credential/network surface added.

## Regression Analysis

Zero regression: 457/457 `packages/create-agent-harness` tests pass (0 failures, 2 pre-existing skips unaffected), `tsc` and full repo build clean, dependent-package test unaffected (1/1).

## ADR

Appended a dated 2026-08-18 section to [`ADR-022`](https://github.com/ruvnet/metaharness/blob/dream/2026-08-18-security-adversarial/docs/adrs/ADR-022-mcp-primitive.md) (the ADR that documents `mcp-scan`'s checks) rather than minting a new ADR number, matching the 2026-08-13/14/17 precedent for a coverage/completion addition to an existing architectural decision.

## Research Gist

No gist-creation tool is available in this session's toolset (matching 2026-08-13 through 08-17) — committed instead: [`docs/dream-cycle/2026-08-18-gist.md`](https://github.com/ruvnet/metaharness/blob/dream/2026-08-18-security-adversarial/docs/dream-cycle/2026-08-18-gist.md) (`GIST=LOCAL`).

## Issue

#209 (full findings, candidate-selection reasoning, adversarial-critique writeup including a self-caught overstatement correction, sbom/policy scan results, competitor review, ledger check/backfill).

## Witness

```
session_commit : 5453c8c990824b54e05f289774e5a8b2cea0a32e
report_sha256  : 007959c64d46f68cfb273142e1b37a175ffd91f53c0c015c1f3be1bfd8517883
witness        : 97587f9cdd8fee182c05ae0f1a47946738195e04ae0126495119f5354c49dc18
```
Verifier procedure is in the gist's Witness section and was independently self-verified (round-trip reproduced programmatically) before this PR was opened.

## Merge Policy

Recommendation: **ACCEPT-WITH-CAVEATS** — small, dual-repro'd (real CLI + deterministic unit) fix to a genuinely live gap in the repo's own security scanner. One disclosed correction along the way: the initial draft overstated the `Bash(*)`/bare-`Bash` half as traceable to a live generator bug at `host-config.ts:45`; an independent critic traced the real call graph, found that path unreachable through any shipped generator today, and the framing was corrected before this PR opened — that half ships as disclosed defense-in-depth, not as a second live-vulnerability claim. Two comparable-class gaps in sibling tools are disclosed as open next steps, not silently fixed under scope pressure: (1) `threat-model` never consults `mcp-scan`'s findings, so the same harness can be HIGH on one tool and "clean" on the other; (2) `harness sbom` has no Cargo/Rust awareness despite `--with-wasm` legitimately shipping Rust deps in scaffolded harnesses.

---
_Generated by [Claude Code](https://claude.ai/code/session_01R1diHYv256aiVNdEEUMfjG)_


ruvnet:
Reviewed, conflict-resolved, and **the live-gap claim verified against a real scaffold** rather than taken from the description.

**The template really does ship it.** `Bash(python *)` is present in both `templates/catalog.def.mjs:310` and `templates/vertical_ai/.claude/settings.json.tmpl:7`, and a real `metaharness ai-bot --template vertical:ai --host claude-code` produces:

```json
["Bash(npx ai-bot*)", "mcp__ai-bot__*", "mcp__experiments__*", "Bash(python *)"]
```

**And the scanner was genuinely blind to it.** Same scaffold, same command, only the scanner source swapped:

| scanner | result |
|---|---|
| `origin/main` (pre-fix) | `HIGH (2 findings, 1 high)` — **no Bash finding at all** |
| this PR | `HIGH (3 findings, 1 high)` — `[MED] Risky shell allow-rule: Bash(python *)` |

So a shipped template grants an unscoped Python interpreter — arbitrary code execution — and `mcp-scan`, the tool whose job is to catch exactly that, said nothing. The old regex listed only `rm|curl|wget|sudo|chmod|ssh`: narrow file/network binaries, no interpreters. That is a real gap, not a hypothetical.

**The honesty in the diff comments is worth noting.** The new `unrestricted-bash-allow` HIGH for `Bash`/`Bash(*)` is explicitly labelled defense-in-depth against a config shape *no current generator emits* — and the comment says so, with the reasoning (web-ui's `claudeSettings()` hardcodes a scoped rule; `host-config.ts`'s `Bash(*)` only reaches shapes mcp-scan doesn't read). Separating "this is live today" from "this is a shape we should refuse anyway" is exactly the distinction that makes a security report trustworthy, and it would have been easy to blur them into one louder claim.

**Tests:** 479 passed / 2 skipped (up from 474). `tsc` clean.

**Fixed before merge — `chore(metaharness): 0.4.12 -> 0.4.13` (`115d9dcc`).** `src/mcp-scan.ts` is shipped code and 0.4.12 is already published. This is the **third** dream PR today to change published code without a bump (avo on #220, flywheel on #258, now this) — three independent instances in one day, which makes the registry preflight gate proposed in #262 clearly worth building rather than continuing to catch these by hand.

**Conflict:** `LEDGER.md` only — merged both row sets, deduplicated by date, 08-13 → 08-29 in order.

PR #215, opened 2026-08-22
feat(avo): gate release claims on exact signed evidence

ruvnet:
## Summary

- add a fail-closed AVO release-claim verifier with an exact protected claim registry
- reject empty inventories, caller-mislabelled mechanism claims, removed governed surfaces, and undeclared AVO performance/frontier marketing
- bind restricted claims to a signed preregistration, exact tag SHA, exact npm tarball/version, unseen 100-task SWE-bench protocol, measured costs, lineage roots, and exact normalized claim semantics
- require externally pinned Ed25519 authorities plus two official graders with distinct keys and organizations
- prepack AVO once in the tag workflow, verify that artifact, and publish those same bytes
- document the boundary in ADR-253 and keep the default manifest to one narrow mechanism claim

## Security invariants

Restricted evidence is accepted only when:

1. the source SHA equals the checked-out release tag;
2. the package name/version and SHA-256 equal the exact tarball passed to npm;
3. registration predates execution and freezes task set, model, reasoning, evaluator, budgets, thresholds, and claim predicates;
4. benchmark totals reconcile across all 300 arm runs and stay within cost/replay/policy gates;
5. benchmark and every grader bind the preregistration, claim-set, artifact, task-set, result, and lineage digests;
6. the trust-policy digest comes from protected repository/environment configuration;
7. all preregistered graders sign, with at least two distinct keys and organizations.

The default release does **not** assert a performance result. No current fixture is accepted as frontier evidence.

## Validation

- `tsc -p packages/avo/tsconfig.json`
- `tsc -p packages/horizon/tsconfig.json`
- `vitest run packages/avo` — 32 passed, 1 live-provider skipped
- `node scripts/avo-claim-gate.mjs --sha 19ab3046b504b7d4cf1cafd338d7f096a4251b86`
- JS syntax checks for both release scripts
- workflow YAML parse for CI and publish
- `git diff --check`
- npm production audit: 0 vulnerabilities
- Ruflo secret scan: 0 secrets

The repository-wide ordered build could not be rerun in this environment because its nested npm invocation requested unavailable network approval. Focused AVO/Horizon builds and tests are green.

## Coordination / residual risk

- No overlapping open PR implements this claim gate.
- PR #214 separately addresses the live agent-context alias; this PR's mechanism statement is intentionally limited to package/API/evidence-schema existence and makes no runtime-safety claim.
- PRs #179/#205 already use ADR-251/252 areas; this PR uses ADR-253.
- Signed lineage booleans remain accountable grader attestations, not local replay of every leaf.
- Repository settings must require CODEOWNER review and protect `AVO_CLAIM_TRUST_POLICY_HASH`.
- Valid restricted evidence needs an external evidence-delivery path keyed to the already-fixed tag SHA; absent that, release fails closed.

Draft: do not merge until release owners review trust-root governance, protected workflow settings, and the stacked runtime hardening.


ruvnet:
**Blocker before review: ADR number collision.**

This PR adds `docs/adrs/ADR-253-avo-release-claim-evidence-gate.md`, but **ADR-253 is already taken on `main`** by `ADR-253-arc-agi-3-chatgpt-harness.md` (landed with #220). Two unrelated ADRs would carry the same number, and `INDEX.md` would gain a second ADR-253 row.

Current high-water mark on `main` is **274**, so this wants **ADR-275**.

This is the **third** such collision I have hit today:

| PR | claimed | already taken by | resolved to |
|---|---|---|---|
| #179 | ADR-251 | `ADR-251-governed-autonomous-variation-runtime` (shipped `@metaharness/avo`) | ADR-273 |
| #258 | ADR-274 | — (was free) | kept 274 |
| **#215** | **ADR-253** | `ADR-253-arc-agi-3-chatgpt-harness` (#220) | **needs 275** |

The pattern is structural, not careless: an ADR number is chosen when a branch is created, and it stays valid only until some *other* branch claiming the same number merges first. With PRs sitting open for days or weeks, collisions are near-certain — and the two ADR-251s were live on `main` simultaneously, which made `ADR-252`'s existing "ADR-251 capability policy" cross-reference ambiguous.

That is the same root cause as the `LEDGER.md` gap documented in #240: **work recorded on a branch that has not merged is invisible to the next author.** Filing that separately as its own issue.

Happy to review the substance once the number is moved — the release-claim gating touches `publish.yml`, `ci.yml` and `CODEOWNERS`, so it deserves a careful pass rather than a rubber stamp, and I would rather do that against a diff that will not need re-numbering afterwards.

ruvnet:
Substance reviewed now that CodeQL is green and the number is moved. **The release hardening is real and well-built.** Merging, with one caveat the PR itself already names.

**`publish.yml` — binds a release to an exact commit.** Three things together:
- `checkout` pinned to `inputs.tag` on dispatch rather than a floating ref
- the tag must match `^v[0-9]+\.[0-9]+\.[0-9]+…$` — no implicit or moving tags
- **the checked-out SHA is compared against the tag's SHA** and fails closed on mismatch

That third check is the one that matters: it closes the window where a tag is repointed between resolution and checkout. `fetch-depth: 0` is correctly included so `git rev-list` can resolve the tag at all.

**`avo-claim-gate.mjs` is hardened where it counts.** Path traversal is guarded, and so is **symlink escape** (`claim-gate symlink escapes repository`) — the case most path checks miss, since a `realpath` check after an `lstat` is the usual gap. Plus a regular-file check, bounded arrays, and fail-closed by default: invoking it bare exits **1**, not 0.

**Two ReDoS paths fixed during review**, both in `normalizeProse` (`51f4e6dd`, `440948ac`). CodeQL reported one witness, I fixed it, and it then reported a *different* payload against the same expression — the label class and the URL class each backtracked independently:

```
payload                   original   label-only fix   both fixed
'[' + N x '[\'             457 ms     0.2 ms          0.1 ms
'[Z](' + N x '[(]('       1081 ms   734 ms  <-- !     0.1 ms
```

Reachable by design: `normalizeProse` runs on claim statements and surface text — the untrusted release-claim prose this gate exists to validate — so a malformed claim could stall the gate itself. Output verified identical on 6/6 markdown cases.

**The caveat, which your own CODEOWNERS comment states:**

> Branch protection must require Code Owner review for these paths; CODEOWNERS alone is not enforcement.

That is exactly right, and worth making concrete: **`main` currently has no branch protection at all** (`GET /branches/main/protection` → 404). So the claim-authority boundary these five entries define is advisory today — anyone with write access can modify `claim-gate.ts` or `publish.yml` without Code Owner review. The PR is not overclaiming, but the protection it depends on has to be switched on separately or the boundary is documentation rather than a control.

avo 55 passed / 1 skipped, runner-coverage gate green, all 6 Node jobs green, CodeQL clean.

PR #220, opened 2026-08-22
feat(arc): add governed AVO harness, ChatGPT adapter, and benchmarks

ruvnet:
## Summary

This pull request adds an independent NVIDIA AVO inspired control architecture for ARC AGI 3 and integrates it with the ChatGPT MCP harness.

It changes the effective agent system, not model weights. ChatGPT proposes candidate plans while MetaHarness validates evidence, selects plans, guards actions, records outcomes, and blocks unsafe or stale execution.

## What changed

1. Added the ARC specific AVO loop with bounded candidate generation, evidence derived scoring, plan lineage, semantic memory, frontier state, guarded execution, blocking supervision, and separately controlled retrodiction.

2. Added explicit experimental arms: DIRECT_ACTOR, AVO_LINEAGE, AVO_MEMORY, AVO_SUPERVISOR_MEMORY, AVO_FULL, and AVO_FULL_RETRODICTION.

3. Added a ChatGPT MCP adapter and frozen actor prompt. The selected ChatGPT model proposes plans through the AVO tool surface and cannot bypass the governed controller. A separate supervisor lane can resolve typed blocking cases but cannot act on the environment.

4. Added durable content addressed AVO checkpoints with configuration binding, compare and swap writes, exact schema validation, ownership checks, and complete receipt coverage.

5. Hardened resume against malformed or rehashed histories, including duplicate candidates, stale observation bases, reordered selections, reordered receipt attribution, extra fields, raw game identity leakage, invalid batch sizes, and nonbijective outcomes.

6. Added a controlled benchmark workspace with direct, direct reflection, and AVO arms; frozen manifests; randomized paired order; complete model usage accounting; clustered confidence intervals; exact sign flip testing; claim gates; and checked evidence artifacts.

7. Added ADR 254 with the experimental design, acceptance rules, rollback plan, evidence boundaries, and the remaining official evaluation gate.

8. Hardened the generic AVO runtime around untrusted JSON, action schemas, evaluator evidence binding, replay integrity, and workspace identity.

## Benchmark evidence

| Evaluation | Baseline | AVO | Result |
| --- | ---: | ---: | --- |
| Synthetic mechanism, 18 paired episodes | Score 0, 8 mean actions | Score 100, 2.333 mean actions | Delta 100, 95 percent CI 100 to 100, exact p 0.015625 |
| Initial online public smoke | Score 3.5714, 80 actions | Score 3.5714, 80 actions | No gain |
| Actor declared clean online rerun | Score 0.3968, first solved level at action 66 | Score 3.2676, first solved level at action 23 | Delta 2.8708, 65.2 percent fewer actions to first solved level |
| Long horizon durability | Not applicable | 6,624 actions | Exact restore of 6,624 receipts, candidates, selections, and outcomes |

The synthetic mechanism gate passes all ten checks and reproduced deterministic evidence hash:

`854514a34bd1ed9679d8ef37b5c16005daa0c19738a96ef35a30589099db91e4`

The durability run reproduced deterministic evidence hash:

`332d1cfd8633ae69e0b5c32663ee4c4eef4cef3f162f4c80761f017108765d09`

## Claim boundary

The synthetic result proves the intended governed planning mechanism on the frozen constructed fixture.

The online result is exploratory only. It covers one public game, one seed, noncompetition mode, and actor declared hygiene. Both arms completed only one of seven levels and exhausted 80 actions.

This pull request does not claim an official ARC AGI 3 score, increased intrinsic model intelligence, NVIDIA implementation equivalence, or general intelligence. The frozen full claim gate remains 25 games and 183 levels in competition mode, mean score improvement of at least 10 points, and a clustered 95 percent lower confidence bound above zero.

## Validation

1. ARC core: 91 of 91 tests passed, including two 6,624 action long horizon cases.

2. ChatGPT adapter: 67 of 67 tests passed.

3. Benchmark package: 11 of 11 tests passed.

4. Generic AVO runtime: 29 tests passed and one credential dependent provider test skipped.

5. Total: 198 tests passed and one skipped.

6. TypeScript checks passed for all four affected workspaces.

7. All checked JSON artifacts parse and their deterministic and report hashes verify.

8. Git diff validation passed with no whitespace errors.

## Review focus

1. Candidate scoring and ablation isolation in `packages/arc-agi-3/src/plan-archive.ts`.

2. Checkpoint verification and receipt binding in `packages/arc-agi-3/src/avo-loop.ts`.

3. ChatGPT runtime attestation and tool boundaries in `packages/arc-agi-3-chatgpt/src/official-factory.ts` and `official-avo-runtime.ts`.

4. Content addressed checkpoint persistence in `packages/arc-agi-3-chatgpt/src/store.ts`.

5. Benchmark validity and claim boundaries in ADR 254 and `packages/arc-agi-3-bench`.

## Acceptance test

Run the full affected test suites and confirm 198 passing tests, then rerun the scripted mechanism benchmark and verify that its deterministic evidence hash remains `854514a34bd1ed9679d8ef37b5c16005daa0c19738a96ef35a30589099db91e4`.


ruvnet:
Reviewed, conflict-resolved, and **ran your acceptance test verbatim**. It passes.

> Run the full affected test suites and confirm 198 passing tests, then rerun the scripted mechanism benchmark and verify that its deterministic evidence hash remains `854514a3…`

**Deterministic evidence hash reproduced exactly:**

```
deterministicEvidenceHash: 854514a34bd1ed9679d8ef37b5c16005daa0c19738a96ef35a30589099db91e4
```

I went further than the stated check and diffed my regenerated report against the committed `results/mechanism-v1.json` field by field. The **only** differences are 54 × `elapsedWallMs`, `generatedAt`, and the `reportHash` that covers them. Every substantive field — episodes, scores, actions, aggregates, comparisons, acceptance — is byte-identical. That is exactly what your `limitations` entry promises ("Observed wall time is recorded per episode but excluded from deterministic acceptance"), and it is why splitting `deterministicEvidenceHash` from `reportHash` is the right design. Two consecutive reruns produced two different `reportHash` values and the same evidence hash.

**Suites, run individually:**

| package | result |
|---|---|
| `arc-agi-3` | 91 passed |
| `arc-agi-3-chatgpt` | 67 passed |
| `arc-agi-3-bench` | 11 passed |
| `avo` | 39 passed, 1 skipped |

**208 passed / 1 skipped, not the 198 in the description.** The whole delta is `avo` (39 vs the stated 29) — `main`'s avo moved after you wrote this; I measured 28 + 1 skipped while reviewing #226 earlier today. Higher than claimed, which is the benign direction, but the number in the PR body is now stale.

**One thing needed fixing before this could merge — `chore(avo): 0.1.4 -> 0.1.5` (`67751052`).** This PR changes `packages/avo/src/operator.ts` (+379/−19) and `types.ts` (+16) while leaving the version at `0.1.4`, which is **already published**. Merging as-is would leave the repo serving different content under a version the registry already has — the same divergence that made `metaharness@0.4.8` unreproducible from `main` (#261), and precisely what the registry preflight gate in #262 is designed to catch. The change is additive (optional `EvaluationBinding`, new failure constants, JSON depth/node/byte bounds, memory retrieval cap), so a patch bump is correct and `create-agent-harness`'s `^0.1.4` still resolves. avo `0.1.5` will be published after merge.

**Conflict:** `INDEX.md` only — kept `main`'s rows plus this PR's new ADR-254 row. ADR-254 is genuinely free (unlike #179, which had collided on 251). The three `arc-agi-3*` packages are all `private: true`, so nothing else publishes.

Also confirms clean against the runner-coverage gate that landed today (#195): 43 workspace visits, up from 42 — the new bench package is reached, not silently unrun.

Claim boundary in the description reads accurately to me: the synthetic mechanism result is the only thing the deterministic hash attests, the online run is explicitly exploratory, and the 25-game/183-level competition gate is still named as outstanding.

ruvnet:
CI caught a flake my local run masked. Fixed in `38e2c82c`.

Three jobs failed — Node 20/ubuntu, Node 20/windows, Node 22/windows — while 8 passed, including every macOS leg and my local run of the same suite:

```
packages/arc-agi-3-bench/__tests__/model-driver.test.ts:115
AssertionError: expected [ "turn_….json" ] to deeply equal [ "turn_….json.partial" ]
```

**The product code is correct; the test was wrong.** `FileBrokerModelDriver` writes `<requestId>.json.partial` and then renames it (`model-driver.ts:446-448`) — a proper atomic write, which is exactly what you want. The test polled `readdir()` until the directory was merely *non-empty*, so on a slower filesystem it could observe the temporary name mid-rename, exit the poll loop, and assert against `.partial`.

Filtering `.partial` out of the poll makes it wait for the settled filename. The logic difference, on a directory in the mid-rename state:

```
OLD poll: exits loop, asserts against ["turn_….json.partial"]  -> fails
NEW poll: keeps polling until the rename settles               -> passes
```

I'm deliberately **not** citing "11/11 passing locally" as evidence the fix works — it passed locally before the fix too, and on 8 of 11 CI jobs. The race is filesystem- and timing-dependent, so a green local run is not the test that matters here; the CI matrix is. Re-running now.

I could not confirm from the annotations whether the two Windows failures share this root cause — their run was still in progress, so only the generic `Process completed with exit code 1` was available rather than the assertion text. If they were a different failure, this push will surface it.

Everything from my previous review still stands: deterministic evidence hash reproduces exactly, 208 passed / 1 skipped across the four suites, `INDEX.md` conflict resolved, and avo bumped 0.1.4 → 0.1.5 because its shipped code changed.

ruvnet:
Recheck split the three failures into two distinct causes, which is why I flagged the Windows ones as unconfirmed rather than assuming one diagnosis covered all of them.

**Node 20/ubuntu now passes** — that was the `.partial` mid-rename race, fixed in `38e2c82c`.

**Both Windows jobs still failed, from an unrelated bug that was never yours.** `scripts/proxy-pin-drift.test.mjs` (added by #178, merged earlier today) asserts the pinned PEM against `/^-----BEGIN PUBLIC KEY-----\n/`. `meta-proxy.ts` is not pinned to `eol=lf` in `.gitattributes`, so a Windows checkout yields a CRLF PEM and the LF-anchored match fails. It was breaking `main` too, not just this PR.

Verified directly rather than inferred — the assertion on this branch vs `main` after the fix:

```
branch head : /^-----BEGIN PUBLIC KEY-----\n/      <- fails on CRLF
main        : /^-----BEGIN PUBLIC KEY-----\r?\n/   <- fixed in 03bba663
```

Production was never affected: the PEM only ever reaches `crypto.createPublicKey()`, is never string-compared or hashed, and `createPublicKey()` accepts CRLF and LF alike (I checked both).

Merged current `main` into this branch (`06d01426`) so it picks up the fix. `proxy-pin-drift` now passes 31 assertions and the runner-coverage gate is clean.

Root cause of that reaching `main` at all is filed as #265: #178 came from a fork, fork PRs report **no checks** on this repo, and I merged it on Linux-only local verification. "No checks reported" is not "checks passed", and treating them as equivalent is what put a Windows break on `main`.

PR #229, opened 2026-08-25
dream(generator-genome): scoreTestConfidence trusted manifest, not real tests (evaluated)

ruvnet:
MetaHarness Dream Cycle, 2026-08-25. Deep surface: `generator-genome` (slot 0). Full detail, receipts, candidate-selection scoring, adversarial-critique writeup, and router/turn-credit scan findings: #228.

## Hypothesis

Given `analyze-repo.ts`'s `testCommands` inference (language/manifest presence only) and `genome-scorers.ts`'s `scoreTestConfidence(profile)` (which credited `testCommands.length > 0` regardless of whether real test files exist), when test-confidence is gated on real filesystem evidence — `hasVerifiedTestFiles`, a `tests`/`__tests__`/`test` directory scan added to `inventory()`, mirroring `score.ts`'s existing `dirExists` pattern — then a repo with a test-capable manifest/CI but zero real test files should score materially lower `test_confidence` (and higher `risk_score`) than an equivalent repo with real test files present, subject to: (a) existing behavior for repos that DO have real tests is unchanged or improves, (b) `score.ts`'s already-correct `scoreTestCoverage` is not regressed, (c) no existing test weakened, (d) diff stays materially under 300 lines, (e) zero LLM/model calls required.

## Candidate

`scoreTestConfidence` credited `profile.testCommands.length > 0` (+0.5) and `profile.hasCi` (+0.3) directly. `testCommands` is built in `analyze-repo.ts` purely from language/manifest detection — a `pyproject.toml`/`requirements.txt` unconditionally pushes `'pytest'`, a `go.mod` unconditionally pushes `'go test ./...'`, a `package.json` with a `scripts.test` key pushes `'npm test'` — none of it checks whether a test file actually exists. A repo with `"test": "echo no tests"` and CI wired, but zero real test files, scored 0.8/1.0.

Fixed: added `RepoProfile.hasVerifiedTestFiles?: boolean`, populated in `inventory()` via a new `dirExists` helper (byte-for-byte the same check as `score.ts`'s) scanning for `tests`/`__tests__`/`test`, read into the profile in `analyzeFiles()`. `scoreTestConfidence` now requires it for the base 0.5 credit; without it, only a small 0.1 "CI exists but unconfirmed" credit applies (0 with no CI either). 4 files, ~170 lines total (well under the 300-line target).

## Evaluation Receipt

Deterministic, non-sampled, real end-to-end repro — no LLM calls used (`LLM_EVAL=available` this session, not needed):
[...]

PR #233, opened 2026-08-26
feat(horizon): evidence carrying termination at final boundary

### LINKED ISSUE #232: Horizon: evidence carrying termination with deterministic replay at COMPLETE boundary
## Finding

Evidence Carrying Termination, arXiv 2608.23623, moves verification to the final completion boundary. A model may propose completion, but a deterministic verifier authorizes it only when every required claim is bound to eligible trace evidence and replay reconstructs the claimed value.

The originating team reports 0 of 66 premature unsupported terminations for ECT versus 40 of 66 for the controller on the frozen held out study, while supported completion was 97 of 132 versus 92 of 132. This is strong measured evidence but not yet independently reproduced in MetaHarness.

## Existing gap

`@metaharness/horizon` already has command authorization, observed tool receipts, resumable checkpoints, halt control, and compaction. `LongHorizonDriver` currently accepts any model `final` step without evidence validation.

## Proposed primitive

Add an optional completion contract:

```ts
CompletionConfig {
  requiredClaims: string[]
  allowedEvidenceRoles?: ('tool' | 'model' | 'summary')[]
  requireAuthorizedToolEvidence?: boolean
  requireSuccessfulToolEvidence?: boolean
  maxEvidenceRefsPerClaim?: number
}
```

A final step may carry a typed `CompletionCertificate`. Each claim references stable transcript event IDs and an optional artifact digest. A trusted `replayCompletion` seam reconstructs the claimed value from referenced evidence.

## Security invariants

1. Completion evidence never grants authority.
2. Tool evidence must be authorized and successful by default.
3. Missing replay fails closed.
4. Stable event IDs are required because compaction can reorder or remove transcript positions.
5. Ambiguous IDs, digest mismatch, missing evidence, or replay mismatch reject completion.
6. Rejected completion becomes a bounded failure signal so the agent can recover but cannot loop forever.
7. Existing behavior remains unchanged when no completion policy is configured.

## Benchmark plan

Baseline: current unconditional `final` acceptance.
Candidate: ECT gate.

Workloads:
1. premature final before required tool evidence
2. failed tool result presented as success
3. denied tool result presented as success
4. stale artifact digest
5. missing required claim
6. replay mismatch
7. recovery after rejected final
8. transcript compaction and reordered evidence

Report unsafe completion rate, supported completion rate, recovery rate, verifier latency, certificate size, and false rejection rate. Include at least 100 deterministic fault injections before model based reproduction.

## Acceptance

Zero unsafe completion on deterministic fault injections, no regression when completion gating is disabled, recovery after a rejected premature final, stable evidence identity across transcript reordering, and less than 5 ms p95 verifier overhead for certificates with at most 32 evidence references per claim.

## ADR

ADR 272 on branch `sota/evidence-carrying-termination`.

## Rollback

Remove the optional completion configuration. No state migration is required.

ruvnet:
Closes #232.

## Summary

Adds an opt in evidence carrying completion gate to `@metaharness/horizon`. A model may propose `final`, but when `DriverConfig.completion` is enabled the driver only accepts completion after every required claim resolves to authorized successful trace evidence and a trusted replay seam reconstructs the claimed value.

## Design

* typed `CompletionCertificate`, `CompletionClaim`, and stable `CompletionEvidenceRef`
* stable tool event IDs derived from the persisted action counter, not transcript positions
* optional artifact digest binding
* successful and authorized tool evidence required by default
* deterministic caller supplied `replayCompletion` trust seam
* missing replay, missing claim, ambiguous event ID, failed or denied tool, digest mismatch, or replay mismatch fails closed
* rejected premature completion is recorded and fed into existing halt control so the agent can recover without unbounded retries
* no behavior change when completion gating is disabled

## Why stable event IDs

Transcript compaction can reorder or remove array positions. Positional evidence references could silently retarget after compaction. Stable event IDs ensure removed evidence fails closed instead of binding to a different event.

## Tests

Adds pure verifier tests for successful replay, transcript reordering, missing claims, missing replay, denied evidence, digest mismatch, duplicate event IDs, and replay mismatch. Adds 128 deterministic unsupported completion fault injections. Adds a WASM gated driver integration test that first attempts unsupported completion, then collects evidence, then completes successfully with a certificate.

## Security

Completion evidence is explicitly not authority. The certificate cannot approve tools, expand command policy, grant capabilities, or mutate the evaluator. Existing `CommandGuard` and tool policy receipts remain authoritative.

## Evidence

Motivated by arXiv 2608.23623, submitted 2026 08 22. The reported frozen study shows 0 of 66 premature unsupported terminations for ECT versus 40 of 66 for its controller, with supported completion 97 of 132 versus 92 of 132. These originating team results are not claimed as reproduced here.

## Scope

Six files, additive API, no dependency change, no persistent migration. ADR 272 documents trust boundary and rollback.

## Acceptance before merge

1. CI and TypeScript build green.
2. Existing Horizon behavior unchanged without a completion config.
3. Deterministic fault injections demonstrate zero unsupported final acceptance.
4. p95 completion verification under 5 ms for certificates bounded to 32 references per claim.
5. No autonomous merge or deployment.


ruvnet:
**Blocker: ADR number collision — the fourth today.**

This PR adds `docs/adrs/ADR-272-evidence-carrying-termination.md`, but **ADR-272 is taken on `main`** by `ADR-272-darwin-numeric-genome-kind.md` (landed with #260). Next free is **275** — though #215 also needs 275, so whichever of the two lands first should take it and the other 276.

Running tally, all found while working the backlog today:

| PR | claimed | taken by | status |
|---|---|---|---|
| #179 | ADR-251 | `ADR-251-governed-autonomous-variation-runtime` | renumbered → 273 |
| #215 | ADR-253 | `ADR-253-arc-agi-3-chatgpt-harness` (#220) | open, needs 275 |
| **#233** | **ADR-272** | `ADR-272-darwin-numeric-genome-kind` (#260) | **open, needs 275/276** |
| #258 | ADR-274 | — | free, kept |

Four PRs, three collisions. Tracked as #267, where I proposed a two-line CI check that would have caught every one of them:

```bash
ls docs/adrs/ADR-*.md | sed -E 's/.*ADR-([0-9]+).*/\1/' | sort | uniq -d
```

**Also needed before merge:** `@metaharness/horizon` is published at **0.2.0** and this PR changes `src/completion.ts`, `src/driver.ts` and `src/index.ts` — shipped code. It needs **0.2.1**, or the registry will serve different content under a version it already has. That is the same divergence as #261, and this would be the sixth same-day instance (avo #220, flywheel #258, mcp-scan #210, genome-scorers #229, darwin #207, now horizon) — see #262 for the registry preflight gate that would catch all of them automatically.

I have not reviewed the substance yet. Happy to, once the number moves — I would rather review a diff that will not need renumbering afterwards, and the ADR is load-bearing here since it defines the termination contract the code implements.

ruvnet:
Substance reviewed now that the number is moved. **The design is sound and the claim-scoping is honest.** One calibration note, no blockers.

**Fail-closed by construction.** Missing certificate, missing replay seam, or zero required claims each return `ok: false` rather than vacuously passing. Defaults are the safe ones: `allowedEvidenceRoles: ['tool']`, `requireAuthorizedToolEvidence: true`, `requireSuccessfulToolEvidence: true`, `maxEvidenceRefsPerClaim: 32`. A caller has to actively widen the contract to weaken it.

**Evidence resolves by stable ID, not transcript position** — `eventsById.get(ref.eventId)`, matching what the interface promises. The subtle case is handled too: when two events share an ID, the reference is **rejected as ambiguous** rather than silently resolving to one of them. I went looking for that to be an untested gap and was wrong — `completion.test.ts` covers it directly (`binds references to stable event ids instead of transcript positions`, plus a fail-closed test naming ambiguous IDs among five modes).

**The doc comment scopes the claim correctly**, which matters more than usual for something named "evidence-carrying termination":

> proves only that declared claims are supported by eligible recorded evidence and reconstruct under the caller's deterministic replay function. It does not prove external truth, safety, or alignment and it cannot grant execution authority.

The ADR keeps the same discipline — the cited study "certifies trace support under declared assumptions, not external truth or general safety."

**The test I most wanted to see is there:** `recovers from premature completion and finishes only after new replayable evidence`. Fail-closed gates are easy to make permanently stuck; that one pins the liveness side.

**Calibration note (not a blocker):** `rejects 128 unsupported completion variants` is really **4 fault classes × 32 repetitions** — `i % 4` selects missing-event / stale-digest / failed-receipt / wrong-replay, and `i` only varies a string suffix. All four classes are genuinely covered and the assertion is real, but the name implies breadth the loop does not have. Renaming to something like "rejects each of 4 unsupported completion fault classes across 128 permutations" would describe it accurately — worth doing precisely because the rest of this PR is careful about not overclaiming.

**Unblocking changes I made** (`4a1d9822`), detailed in the commit: ADR-272 → **275** (272 was taken by #260; the merge had left two ADR-272 files and *no* INDEX row for this one), `## Status` → the bold form used by 237/239 ADRs, and `@metaharness/horizon` 0.2.0 → **0.2.1** since `src/{completion,driver,index}.ts` are shipped code.

horizon: 8 passed / 21 skipped — the skips are the pre-existing `describe.skipIf(!wasmOk)` suites that need a Rust toolchain, not anything this PR introduced. `tsc` clean, runner-coverage gate green.

